PR #30, opened 2021-06-30
WTR Component Tests

mabry1985:


mabry1985:
@himerus I used the official Lit starter as a reference for this feature, but I wanted to only bring in the minimal necessary dependencies in functionality. That being said Polymer is using a few test runners and helpers that may or may not be useful for us, like Chai, Karma, or Mocha runners.

You may find something [here](https://github.com/PolymerLabs/lit-element-starter-ts) that I didn't.

himerus (approved):

PR #38, opened 2021-07-08
Feature Draft: Docker local development environment.

damontgomery:
I sometimes (depends on the project) like to use Docker environments for local development.

This would be completely optional and the environments themselves are very simple. Would others find this useful?

Tested
```
yarn install
yarn start
yarn lint
yarn fmt
```

They all seem to work from the container. `yarn install` is probably a little slow since it isn't using NFS or anything.

This probably cannot be committed as-is, particularly since it will conflict with #36.


damontgomery:
Closing to continue discussions around development environments and avoid confusion.

PR #46, opened 2021-07-10
Accordion component

### LINKED ISSUE #37: Accordion Component
## Requirement
- Create an accordion component that is accessible called `outline-accordion`. 
- Accordion should have optional flag to have only one open, or many open. 

## Optional features
- I can set the header to use h1, h2, etc using the `outline-header` component
- I can set link to an accordion using an Anchor link
  - I can copy the link using the browser

### Example(s)
https://designsystem.digital.gov/components/accordion/
Cursor_and_Accordion___U_S__Web_Design_System__USWDS_

### Accessibility Guidelines
[W3.org](https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html)
[A11y Style Guide](https://a11y-style-guide.com/style-guide/section-navigation.html#kssref-navigation-accordion)

LeeMellon:
Accordion Mk I:
 - Basic wireframe styles for now. 
 - Keyboard accessible with tab/up/down arrow controls.
 - Tested in chrome, FF, edge. IE11 not liking our storybook setup)
 - Single open panel at a time optional via singlePanel prop.
 
Cursor_and_Molecules___Accordion_-_Accordion_⋅_Storybook

 

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/46"></a>


src/components/base/outline-accordion/outline-accordion.css
```diff
@@ -0,0 +1,14 @@
+.accordion-buttton {
```
danmuzyka:
```suggestion
.accordion-button {
```

Nitpick: `buttton` should be spelled with two `t`s.

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  render(): TemplateResult {
+    return html`
+      <div class="accordion">
+        ${this.data.map(
+          (panel, index) =>
+            html` <div class="accordion-panel">
+              <h4 class="accordion-heading">
+                <button
+                  class="accordion-buttton"
```
danmuzyka:
```suggestion
                  class="accordion-button"
```

Fix the spelling error on `buttton` (change to `button`) here and in the CSS (comment already left in the CSS file).

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,116 @@
+import { html, TemplateResult, CSSResultGroup } from 'lit';
+import { customElement, property, state } from 'lit/decorators.js';
+import { OutlineElement } from '../outline-element/outline-element';
+import { unsafeHTML } from 'lit/directives/unsafe-html.js';
+import componentStyles from './outline-accordion.css.lit';
```
danmuzyka, line 5: This question might be the result of my lack of experience with Outline and Lit, but I'm wondering why in this import line, you added a `.lit` file extension to the CSS file, when the CSS file itself just has a `.css` file extension. Is this a Lit coding convention, which causes some CSS post-processing to be triggered?

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,116 @@
...
+  static styles: CSSResultGroup = [componentStyles];
+
+  /**
+   * The data aray of <AccordionPanel> objects.
+   * AccordionPanel.content is of indeterminate type, and is passed into the panel <slot>.
+   */
+  @property({ type: Array }) data!: AccordionPanel[];
+  /**
+   * Set to 'true' to create an accordion that can only have one panel open ata a time.
+   */
```
danmuzyka: I LOVE all of the code comments in here!

Styling nitpick: I assume that the coding standards for Outline require a blank line above multiline comments like this? Did this pass linting? Also, `ata` should be `at a`.

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,116 @@
...
+                  aria-controls="${this.seed}-${index}"
+                  @click=${() => this.toggleHidden(`${this.seed}-${index}`)}
+                >
+                  ${panel.heading}
+                </button>
+              </h4>
+              <div
+                class="accordion-content"
+                id="${this.seed}-${index}"
+                .hidden=${!this.active.includes(`${this.seed}-${index}`)}
```
danmuzyka, line 66: Is the `.hidden` attribute a Lit convention? I'm curious about the approach of prefixing the attribute with a dot.

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,116 @@
...
+              >
+                ${unsafeHTML(`${panel.content}`)}
+              </div>
+            </div>`
+        )}
+      </div>
+    `;
+  }
+
+  setActive(contentId: string) {
```
danmuzyka, line 80: Is it not standard practice to include a docblock comment for class methods in Outline?

danmuzyka (changes requested):

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,116 @@
...
+  /**
+   * Set to 'true' to create an accordion that can only have one panel open ata a time.
+   */
+  @property({ type: Boolean }) singlePanel = false;
+
+  /**
+   * ID of the most recently triggered panel.
+   * If the id is new, sets focus on the panel content div to facilitate screen readers.
+   */
+  @state() focused: string;
```
shaal:
```suggestion
```
A11y:
Because this component is using semantic markup (ie. button to open/close accordion panels),
it shouldn't automatically move the focus to a different element.

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,116 @@
...
+      return (this.active = this.active.filter(item => item !== contentId));
+    }
+    return this.active.push(contentId);
+  }
+
+  setFocused(contentId: string) {
+    this.focused === contentId
+      ? (this.focused = '')
+      : (this.focused = contentId);
+  }
```
shaal:
```suggestion
```
A11y:
Because this component is using semantic markup (ie. button to open/close accordion panels),
it shouldn't automatically move the focus to a different element.

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,116 @@
...
+   * Type '() => boolean' is not assignable to 'boolean'lit-plugin(no-incompatible-type-binding)(2304)
+   */
+
+  isActive(contentId: string): boolean {
+    return this.active.includes(contentId);
+  }
+
+  toggleHidden(contentId: string) {
+    this.setActive(contentId);
+    this.setFocused(contentId);
```
shaal:
```suggestion
```
A11y:
Because this component is using semantic markup (ie. button to open/close accordion panels),
it shouldn't automatically move the focus to a different element.

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,116 @@
...
+    this.requestUpdate();
+  }
+
+  updated() {
+    const focusEle = this.shadowRoot?.getElementById(this.focused);
+    focusEle?.focus();
+  }
```
shaal:
```suggestion
```
A11y:
Because this component is using semantic markup (ie. button to open/close accordion panels),
it shouldn't automatically move the focus to a different element.

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,116 @@
...
+                  @click=${() => this.toggleHidden(`${this.seed}-${index}`)}
+                >
+                  ${panel.heading}
+                </button>
+              </h4>
+              <div
+                class="accordion-content"
+                id="${this.seed}-${index}"
+                .hidden=${!this.active.includes(`${this.seed}-${index}`)}
+                tabindex="0"
```
shaal:
```suggestion
```
A11y:
No need to add `tabindex`

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,116 @@
...
+                <button
+                  class="accordion-buttton"
+                  aria-expanded=${this.active.includes(`${this.seed}-${index}`)}
+                  aria-controls="${this.seed}-${index}"
+                  @click=${() => this.toggleHidden(`${this.seed}-${index}`)}
+                >
+                  ${panel.heading}
+                </button>
+              </h4>
+              <div
```
shaal, line 61:
```suggestion
              <div
                role="region"
                aria-labelledby="[id_of_button_here]"
```
A11y:
The section of content of accordion item should have `role="region"`,
That section should also have `aria-labelledby=` and the id should be of the button that operates (and describes) it.

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,116 @@
...
+  seed = Math.floor(Math.random() * 10000);
+
+  render(): TemplateResult {
+    return html`
+      <div class="accordion">
+        ${this.data.map(
+          (panel, index) =>
+            html` <div class="accordion-panel">
+              <h4 class="accordion-heading">
+                <button
```
shaal, line 50: Button is missing ID (to be referred by `aria-labelledby` on a comment below)

shaal (changes requested):
@LeeMellon awesome work!
I added accessibility suggestions.

It would be great to add a symbol that displays open/close for each of the accordion buttons.
(That symbol should be hidden from screen-readers, since they use `aria-expanded` attribute to identify the status)

It would be great to add additional support for keyboard navigation keys (`home`/`end`/up/down/left/right keys)

src/components/base/outline-accordion/outline-accordion.ts
mabry1985: Yeah this has to do with how we are processing the CSS, it converts the css into a css template literal that can be read from the component

src/components/base/outline-accordion/outline-accordion.ts
mabry1985: Hidden is just an html attribute. The period is a lit-html convention that allows you to directly bind data to that attribute. Here is a pretty good write up on that and a couple other nuances https://terodox.tech/lit-html-part-2/

src/components/base/outline-accordion/outline-accordion.css
LeeMellon: Not nitpick at all That's me being sloppy dev. Good catch!

src/components/base/outline-accordion/outline-accordion.ts
LeeMellon: No that's a good question that I'm not fully clear on myself. When we build SB/run yarn start we call ./scripts/lit-styles.js which generates a "standard" css (**.css.lit) file from the tailwind styles in the **.css file. So we're telling the component to look into that generated file for the styles. The why of all of this is a question better directed to @himerus.

src/components/base/outline-accordion/outline-accordion.ts
LeeMellon: Yeah it passed linting. @mabry1985 gets a lot of the credit as I'm straight mimicking his comment style because it does look good.

src/components/base/outline-accordion/outline-accordion.ts
LeeMellon: Still a bit fuzzy on this myself.
https://lit.dev/docs/templates/expressions/#:~:text=Property%20expressions-,%23,Permalink%20to%20Property%20expressions%20%7B%23property-expressions%7D,-You%20can%20set

src/components/base/outline-accordion/outline-accordion.ts
LeeMellon: I don't think there's a standard anything in Outline yet, but it's not a bad time to start!

src/components/base/outline-accordion/outline-accordion.ts
damontgomery: Thanks for the links. :)

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,143 @@
...
+@customElement('outline-accordion')
+export class OutlineAccordion extends OutlineElement {
+  static styles: CSSResultGroup = [componentStyles];
+
+  /**
+   * The panels aray of <AccordionPanel> objects.
+   * AccordionPanel.content is of indeterminate type, and is passed into the panel <slot>.
+   */
+  @property({ type: Array }) panels!: AccordionPanel[];
+  /**
```
danmuzyka: Code formatting nitpick: Should there be be a blank line before the comment? Or is that not part of the coding style here? Looking forward to the style guide doc that Josh is going to start, to help us clarify the expectations!

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,143 @@
...
+            </div>`
+        )}
+      </div>
+    `;
+  }
+
+  /**
+   * Takes the element id of content <div>
+   * to maintain state list of active/open panels.
+   */
```
danmuzyka, line 79: Thanks for adding this comment! Thoughts on using `@param` and `@return` in the comment for each method?

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,143 @@
...
+    this.requestUpdate();
+  }
+
+  /**
+   *
+   * @param event
+   * @returns void
+   * Adds up/down arrow keyboard navigation.
+   * TODO: try to refactor to something simpler.
+   */
```
danmuzyka: I LOVE that you added a description and `@param` and `@returns` here. I usually expect to see the description above the `@` metadata, though. Any concerns about changing that?

src/components/base/outline-accordion/outline-accordion.ts
```diff
@@ -0,0 +1,143 @@
...
+    }
+    return this.active.push(contentId);
+  }
+
+  /**
+   *
+   * Dupliacted/un-utilized in the template due to lit-plugin error
+   * Type '() => boolean' is not assignable to 'boolean'lit-plugin(no-incompatible-type-binding)(2304)
+   */
+
```
danmuzyka: Nitpick: I would expect not to see a blank line between a comment and its corresponding method. I also would not expect to see the more or less empty part of the comment block on line 93. The `@param` and `@return` metadata would be great here. 😃

danmuzyka (dismissed):

src/components/base/outline-accordion/outline-accordion.ts
mabry1985: @danmuzyka can you clarify what you mean by the having the description above the meta data?
`event`
`@param`

Like that?

src/components/base/outline-accordion/outline-accordion.ts
himerus: I would definitely be in favor of the blank line for separation.

LeeMellon: This was how prettier formatted it. I'll see if it will let me make the change.

mabry1985:
@LeeMellon How are we looking on this component? Anything I can do to help?

LeeMellon:
I think we're looking fine. I've been waiting, a week for another look from people. 
https://phase2.slack.com/archives/C01RRSPB62F/p1626305712158600

shaal (approved):
LGTM! 
Suggested follow up - add a clickable icon that will display the status of each accordion panel.

mabry1985 (approved):
This is looking good to me. Let's bring it in and create an issue for adding the icon button and doing some styling.

danmuzyka (approved):
I think we may want to make some changes based on the code style guide (e.g., related to code comments), but that could be done as a follow-up ticket. Approving this in its current state to prevent blocking this any further.

Issue #45, opened 2021-07-09
Modal Component


LeeMellon:
## Requirements

Create a basic modal component that:

- Can be closed via button or keyboard (esc key).
- Has small, medium, full screen sizes.
- At this stage passes all content into a slot.
- Meets accessibility standards.

### Examples
https://designsystem.digital.gov/components/modal/
![Modal___U_S__Web_Design_System__USWDS_](https://user-images.githubusercontent.com/11083407/125133930-6997e300-e0bb-11eb-93f4-f21f876f4777.png)

### Accessibility Guidelines
[W3.org](https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html)

damontgomery:
This looks like it works pretty well. https://patternflyelements.org/components/modal/

damontgomery:
#49 is ready for review.

PR #49, opened 2021-07-12
Feature: #45 Modal.

damontgomery:
# Description

Adds a modal. See #45 

# Testing notes
- Visit http://localhost:6006/?path=/story/molecules-modal--small and similar examples.

# Example

modal_canvas

## Documentation

modal_documentation

# Notes

Took concepts from https://micromodal.vercel.app and https://patternflyelements.org/components/modal

The controls for slots in Storybook are weird and this documentation works, but is awkward. Open to suggestions!

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/49"></a>

### DIFF
--- a/src/components/base/outline-modal/outline-modal.css
+++ b/src/components/base/outline-modal/outline-modal.css
@@ -10,7 +10,7 @@
 }
 
 #container {
-  @apply overflow-y-auto rounded-sm max-h-screen bg-neutral-white max-w-full md:max-w-quarter-screen-w;
+  @apply overflow-y-auto rounded-sm h-screen bg-neutral-white max-w-full md:max-w-quarter-screen-w md:h-auto;
   min-width: 320px;
 
   #overlay.medium & {
--- a/src/components/base/outline-modal/outline-modal.css
+++ b/src/components/base/outline-modal/outline-modal.css
@@ -1,0 +1,44 @@
+* {
+  box-sizing: border-box;
+}
+
+#overlay {
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background: rgba(0,0,0,0.6);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+#container {
+  background-color: #fff;
+  min-width: 320px;
+  max-width: 500px;
+  max-height: 100vh;
+  border-radius: 4px;
+  overflow-y: auto;
+}
+
+#header {
+  position: relative;
+  padding: 1.62rem 1.62rem 0 1.62rem;
+}
+
+#close {
+  background: transparent;
+  border: 0;
+  position: absolute;
+  top: 0.62rem;
+  right: 0.62rem;
+}
+
+/* We need a double slash on the unicode because of templates later. */
+#close:before { content: "\\2715"; }
+
+#main {
+  padding: 1.62rem;
+}
--- a/src/components/base/outline-modal/outline-modal.css
+++ b/src/components/base/outline-modal/outline-modal.css
@@ -10,9 +10,16 @@
 }
 
 #container {
-  @apply overflow-y-auto rounded-sm max-h-screen bg-neutral-white;
+  @apply overflow-y-auto rounded-sm max-h-screen bg-neutral-white max-w-full md:max-w-quarter-screen-w;
   min-width: 320px;
-  max-width: 500px;
+
+  #overlay.medium & {
+    @apply md:max-w-half-screen-w;
+  }
+
+  #overlay.full-screen & {
+    @apply w-full-screen-w max-w-full-screen-w h-full-screen-h md:w-full-screen-w md:max-w-full-screen-w;
+  }
 }
 
 #header {
--- a/src/components/base/outline-modal/outline-modal.css
+++ b/src/components/base/outline-modal/outline-modal.css
@@ -38,3 +38,7 @@
 #main {
   @apply p-6;
 }
+
+#accessibility-description {
+  @apply hidden;
+}
--- a/src/components/base/outline-modal/outline-modal.css
+++ b/src/components/base/outline-modal/outline-modal.css
@@ -1,7 +1,3 @@
-* {
-  @apply box-border;
-}
-
 #overlay {
   @apply fixed top-0 left-0 right-0 bottom-0 z-50 flex items-center justify-center bg-opacity-60 bg-neutral-black;
 
--- a/src/components/base/outline-modal/outline-modal.css
+++ b/src/components/base/outline-modal/outline-modal.css
@@ -12,6 +12,7 @@
   display: flex;
   justify-content: center;
   align-items: center;
+  z-index: 99999;
 }
 
 #container {
--- a/src/components/base/outline-modal/outline-modal.css
+++ b/src/components/base/outline-modal/outline-modal.css
@@ -3,11 +3,14 @@
 }
 
 #overlay {
-  @apply fixed top-0 left-0 right-0 bottom-0 z-50 flex items-center justify-center bg-black bg-opacity-6;
+  @apply fixed top-0 left-0 right-0 bottom-0 z-50 flex items-center justify-center bg-opacity-60 bg-neutral-black;
+
+  /* Tailwind background opacity with customizations wasn't working.\`bg-opacity-60 bg-neutral-black\` */
+  background-color: rgb(0, 0, 0, var(--tw-bg-opacity));
 }
 
 #container {
-  @apply overflow-y-auto bg-white rounded-sm min-h-screen;
+  @apply overflow-y-auto rounded-sm max-h-screen bg-neutral-white;
   min-width: 320px;
   max-width: 500px;
 }
@@ -17,7 +20,7 @@
 }
 
 #close {
-  @apply absolute bg-transparent border-0 top-2.5 right-2.5;
+  @apply absolute border-0 top-2.5 right-2.5 bg-neutral-transparent;
 }
 
 /* We need a double slash on the unicode because of templates later. */
--- a/src/components/base/outline-modal/outline-modal.css
+++ b/src/components/base/outline-modal/outline-modal.css
@@ -1,40 +1,23 @@
 * {
-  box-sizing: border-box;
+  @apply box-border;
 }
 
 #overlay {
-  position: fixed;
-  top: 0;
-  left: 0;
-  right: 0;
-  bottom: 0;
-  background: rgba(0, 0, 0, 0.6);
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  z-index: 99999;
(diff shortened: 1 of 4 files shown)


src/components/base/outline-modal/outline-modal.stories.ts
```diff
@@ -0,0 +1,106 @@
...
+
+const Template = ({
+  triggerSlot,
+  headerSlot,
+  defaultSlot,
+  size,
+}): TemplateResult => {
+  return html`
+    <outline-modal size="${ifDefined(size)}">
+      ${ifDefined(triggerSlot)} ${ifDefined(headerSlot)}
```
damontgomery, line 180: Prettier?? Not sure why it doesn't want this on a separate line. :(

damontgomery:
I worked on documentation notes to describe functionality and usage.

Added events so others can detect what's happening and respond to changes.

Switched to using Tailwind classes in most cases.

Still need to add styles to different sizes of the modal (small, medium, full-screen).

damontgomery:
I added the styles for different sizes. Going to review this later today and see if there are other things I need to add.

damontgomery:
Going to review the W3 requirements, https://www.w3.org/TR/wai-aria-practices-1.1/examples/dialog-modal/dialog.html.

src/components/base/outline-modal/outline-modal.css
```diff
@@ -0,0 +1,40 @@
+* {
+  @apply box-border;
```
LeeMellon: Not 100% sure this is needed, but I could be wrong.

src/components/base/outline-modal/outline-modal.ts
```diff
@@ -0,0 +1,134 @@
...
+  private _overlayTemplate(): TemplateResult {
+    let template = html``;
+
+    if (this.isOpen) {
+      template = html`
+        <div
+          id="overlay"
+          tabindex="-1"
+          class="${this.size}"
+          @click="${this._handleModalClose}"
```
LeeMellon: Wondering if it might be worth it to make this conditional? In the case where you may want to force the user to take a specific action before closing the modal, rather than just being able to click the overlay to close it.

LeeMellon (approved):
Looks fantastic! Two comments that probably speak more to my ignorance than anything. Would still be okay merging this.

src/components/base/outline-modal/outline-modal.css
damontgomery: Thanks for taking a look!

I don't know either, but I saw it in a best-practices. Maybe it's part of the global styles? I can try removing.

src/components/base/outline-modal/outline-modal.ts
damontgomery: It's a good question. I don't have strong opinions. Maybe we can discuss as a group?

damontgomery:
I tried to add a bunch of accessibility features around focus. Still need to trap the focus within the modal.

It also surfaced an issue where I wasn't able to focus our `outline-button` element. Not sure why.

src/components/base/outline-modal/outline-modal.css
damontgomery: It didn't seem to do anything here, so I removed it. Thanks again.

src/components/base/outline-modal/outline-modal.ts
```diff
@@ -0,0 +1,263 @@
...
+import { OutlineElement } from '../outline-element/outline-element';
+import { ifDefined } from 'lit/directives/if-defined';
+
+export type ModalSize = 'small' | 'medium' | 'full-screen';
+
+export const modalSizes: ModalSize[] = ['small', 'medium', 'full-screen'];
+
+// See https://stackoverflow.com/questions/1599660/which-html-elements-can-receive-focus.
+// @todo make this re-usable across components?
+const focusableElementSelector = `
```
LeeMellon, line 13: I don't know if you need this. This is describing default behaviors, and in the case of the modal I think you can provide prop that takes an element id that gets focus when the modal opens, and that defaults to the "#main" div wrapping your default slot.

src/components/base/outline-modal/outline-modal.css
LeeMellon: Yeah I think that is part of global styles.

src/components/base/outline-modal/outline-modal.ts
```diff
@@ -0,0 +1,263 @@
...
+        break;
+      case 'keydown':
+        if ('key' in event && event.key === 'Enter') {
+          shouldOpen = true;
+          // This prevents a focused element from also triggering.
+          // For example, the modal opens and the "accept" button is focused and then triggered and the modal closes.
+          event.preventDefault();
+          break;
+        }
+    }
```
damontgomery: I think I should probably split this into two handlers. One for the MouseEvent and one for the KeyboardEvent.

The code is a bit difficult to read as-is, because it takes two different inputs, spends most of the time figuring out which of those we are looking at, and then does something very simple.

I'll try to give this a shot on Monday.

src/components/base/outline-modal/outline-modal.ts
```diff
@@ -0,0 +1,263 @@
...
+    let template = html``;
+
+    if (this.isOpen) {
+      template = html`
+        <div
+          id="overlay"
+          tabindex="-1"
+          class="${this.size}"
+          @click="${this._handleModalClose}"
+          @keydown="${this._handleOverlayKeyup}"
```
damontgomery: This needs to be updated. When I switched from `keyup` to `keydown`, I forgot to rename the method. :(

I'll try to tackle that on Monday.

src/components/base/outline-modal/outline-modal.ts
```diff
@@ -0,0 +1,263 @@
...
+              this._getAccessibilityDescriptionId()
+            )}"
+          >
+            <div id="header">
+              <slot id="title" name="outline-modal--header"></slot>
+              <button
+                id="close"
+                aria-label="Close modal"
+                @click="${this._handleModalClose}"
+                @keydown="${this._handleModalCloseKeyup}"
```
damontgomery: And mismatch between `keyup` and `keydown` here too. :(

src/components/base/outline-modal/outline-modal.ts
```diff
@@ -0,0 +1,251 @@
...
+
+    let lastFocusableElement: HTMLElement = this.closeElement;
+
+    const focusableElements: NodeListOf<HTMLElement> = this.querySelectorAll(
+      focusableElementSelector
+    );
+
+    if (focusableElements.length > 0) {
+      lastFocusableElement = focusableElements[focusableElements.length - 1];
+    }
```
damontgomery: Would this make more sense?

```typescript
const focusableElements: NodeListOf<HTMLElement> = this.querySelectorAll(focusableElementSelector);

const lastFocusableElement: HTMLElement = focusableElements[focusableElements.length - 1] ?? this.closeElement;
```

This avoids re-assigning to `lastFocusableElement`.

src/components/base/outline-modal/outline-modal.ts
```diff
@@ -0,0 +1,251 @@
...
+      const automaticallySelectedElementToFocus = this.querySelector(
+        focusableElementSelector
+      ) as HTMLElement | null;
+
+      if (automaticallySelectedElementToFocus !== null) {
+        elementToFocus = automaticallySelectedElementToFocus;
+      }
+    }
+
+    elementToFocus.focus();
```
damontgomery: Does this make more sense? It avoids changing a variables values.

```typescript
const defaultElement: HTMLElement = this.closeElement;

const attributeDefinedElement = (this.elementToFocusSelector !== undefined) ? this.querySelector(this.elementToFocusSelector) as HTMLElement | null : null;

const automaticallySelectedElement = this.querySelector(focusableElementSelector) as HTMLElement | null;

const element = attributeDefinedElement ?? automaticallySelectedElement ?? defaultElement;

element.focus();
```

src/components/base/outline-modal/outline-modal.ts
himerus: This would be a nice feature. Another prop of `canClose` or something, but this would require additional work too so we define what the confirmation button/link/text is/does/etc. 

If it's not "easy", I'd say we add this as a new feature request and iterate.

damontgomery: Thanks for taking a look. https://designsystem.digital.gov/components/modal/ has an example of this we can use as a starting point. For now, I think making the close button optional and requiring the developer to configure the actions is easiest / most flexible and we could look at other options later?

I can probably take a look today.

damontgomery: I've added this as an attribute `shouldForceAction`. There is a story for this in the latest version. Should remove the close features and adjusts the keyboard navigation for focus.

damontgomery:
Added an automated test that checks that the slots are filled when the modal is rendered, opens, and closes.

Also added the `shouldForceAction` attribute so we can force the user to take an action if needed.

src/components/base/outline-modal/outline-modal.ts
```diff
@@ -0,0 +1,282 @@
...
+                : undefined
+            )}"
+          >
+            <div id="header">
+              <slot
+                id="title"
+                name="outline-modal--header"
+                @slotchange="${this._handleSlotChange}"
+              ></slot>
+              ${this.shouldForceAction
```
mabry1985, line 125: Can we place `this.shouldForceAction` in it's own template function? See [here](https://github.com/phase2/outline/pull/60#discussion_r679571323) for examples

src/components/base/outline-modal/outline-modal.ts
```diff
@@ -0,0 +1,282 @@
...
+          tabindex="-1"
+          class="${this.size}"
+          @click="${this._handleOverlayClick}"
+          @keydown="${this._handleOverlayKeydown}"
+        >
+          <div
+            id="container"
+            role="dialog"
+            aria-modal="true"
+            aria-labelledby="${ifDefined(
```
mabry1985, line 110: Let's move this conditional logic for `aria-labelledby` and `aria-described` into variables at the beginning of the render function to clean up this template a bit:
```
render() {
const ariaDesc = this._hasAccessibilityDescriptionSlot ? 'accessibility-description' : null
const ariaLabel = this._hasHeaderSlot ? 'header' : null
 ....
 <div aria-labeledby="${ifDefined(ariaLabel)} aria-describedby="${ifDefined(ariaDesc}">
....
}
```

mabry1985 (changes requested):
I think this is looking pretty good. I appreciate all the great accessibility work here!

I've suggested a couple changes to refactor a bit. 

I would also like to see some more styling done on this component, but lets make an issue for that and I can address that in a separate PR

mabry1985 (approved):
I think this is looking good, there were a couple changes that were discussed previously that I resolved based off of our meeting earlier and I think this is ready to come in and be iterated on.

PR #54, opened 2021-07-14
docs: outline-link storybook help.

damontgomery:
Add documentation around the `outline-link` component to try to give an idea of how it is used.

- code sample
- description

## Testing
- http://localhost:6006/?path=/docs/atoms-link--link

Can someone describe why this would be used instead of an `a` element?

## Example

link_docs

### Before

link_before

### DIFF
--- a/src/components/base/outline-link/outline-link.stories.ts
+++ b/src/components/base/outline-link/outline-link.stories.ts
@@ -33,6 +33,32 @@ export default {
     linkTarget: '_blank',
     linkText: 'Sample Link',
   },
+  parameters: {
+    docs: {
+      description: {
+        component: `
+This component renders a link as an \`a\` element.
+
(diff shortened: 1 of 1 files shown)


labbydev:
> Can someone describe why this would be used instead of an `a` element?

Most projects will require some sort of link web component in their library. This will be a standard starting point for that. We'll use the web components as opposed to `a` tags to control styling and formatting and prevent styling from leaking out or in the components to other elements on the page. 

Should we also include active and inactive states?

himerus:
Agreed the demos of the links could be better. 
I think we can quote some of your words @labbydev for this section. 

Also need to better demonstrate the "why use slotted" the way Drupal does in the documentation as well, This can be a good starter though and I can ticket updating the link component.

Issue #41, opened 2021-07-09
Alert Component


LeeMellon:
## Requirements

Create an alert component that:

- Has large and small size options.
- Has an optional icon (can be hard coded for now until icon component is created/refined).
- Optional: Has basic style options like 'Warning", "Error" "Success" "Basic" (see below).

### Example 

- https://designsystem.digital.gov/components/alert/ 

![Alert___U_S__Web_Design_System__USWDS_](https://user-images.githubusercontent.com/11083407/125127028-d3f75600-e0b0-11eb-99f3-7fa4eda61fcb.png)

### Accessibility Guidelines
[W3.org](https://www.w3.org/TR/wai-aria-practices-1.1/examples/alert/alert.html)

damontgomery:
Going to give this a shot.

damontgomery:
Opened https://github.com/phase2/outline/pull/60.

PR #60, opened 2021-07-16
Feature: #41 Alert.

damontgomery:
# Description

Adds an alert. See #41.

# Testing notes
- Visit http://localhost:6006/?path=/docs/molecules-alert--information and similar examples.

# Example

alert

# Notes

Took styling and types mostly from https://github.com/phase2/outline/issues/41.

This doesn't render an icon yet since there are still some open questions around that.

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/60"></a>

### DIFF
--- a/src/components/base/outline-alert/outline-alert.stories.ts
+++ b/src/components/base/outline-alert/outline-alert.stories.ts
@@ -38,6 +38,10 @@ export default {
       description: '(not yet implemented) Should we show the icon',
       control: { type: 'boolean' },
     },
+    isInteractive: {
+      description: 'Is there an interaction in the alert, such as a button.',
+      control: { type: 'boolean' },
+    },
   },
   parameters: {
     docs: {
@@ -52,6 +56,10 @@ You can set the type of alert with \`statusType\`.
 You can use a smaller alert with \`size\` of \`small\`.
 
 You can remove the icon with \`shouldShowIcon\` set to \`false\`.
+
+## Accessibility
+
+If the alert has an interaction, you should indicate this with \`isInteractive\` set to \`true\`.
 `,
       },
       source: {
@@ -61,6 +69,7 @@ You can remove the icon with \`shouldShowIcon\` set to \`false\`.
   statusType="{{ statusType }}"
   size="{{ size }}"
   shouldShowIcon="{{ shouldShowIcon }}"
+  isInteractive="{{ isInteractive }}"
 >
   <span slot="outline-alert--header">{{ headerSlot }}</span>
   {{ defaultSlot }}
@@ -77,12 +86,14 @@ const Template = ({
   statusType,
   size,
   shouldShowIcon,
+  isInteractive,
 }): TemplateResult => {
   return html`
     <outline-alert
       statusType="${ifDefined(statusType)}"
       size="${ifDefined(size)}"
       shouldShowIcon="${ifDefined(shouldShowIcon)}"
+      isInteractive="${ifDefined(isInteractive)}"
     >
       ${ifDefined(headerSlot)} ${ifDefined(defaultSlot)}
     </outline-alert>
@@ -135,10 +146,11 @@ NoIcon.args = {
   shouldShowIcon: false,
 };
 
-export const MessageWithLink = Template.bind({});
-MessageWithLink.args = {
+export const InteractiveAlert = Template.bind({});
+InteractiveAlert.args = {
   defaultSlot: html`
-    Here is an alert with a link.
+    Here is an alert with an interaction.
     <outline-link linkhref="#">Click here</outline-link>
   `,
+  isInteractive: true,
 };
--- a/src/components/base/outline-alert/outline-alert.stories.ts
+++ b/src/components/base/outline-alert/outline-alert.stories.ts
@@ -91,31 +91,31 @@ const Template = ({
 
 export const Information = Template.bind({});
 Information.args = {
-  defaultSlot: html` Here is an informational message. `,
+  defaultSlot: html`Here is an informational message.`,
   statusType: 'information',
 };
 
 export const Warning = Template.bind({});
 Warning.args = {
-  defaultSlot: html` Here is a warning message. `,
+  defaultSlot: html`Here is a warning message.`,
   statusType: 'warning',
 };
 
 export const Error = Template.bind({});
 Error.args = {
-  defaultSlot: html` Here is an error message. `,
+  defaultSlot: html`Here is an error message.`,
   statusType: 'error',
 };
 
 export const Success = Template.bind({});
 Success.args = {
-  defaultSlot: html` Here is a success message. `,
+  defaultSlot: html`Here is a success message.`,
   statusType: 'success',
 };
 
 export const Small = Template.bind({});
 Small.args = {
-  defaultSlot: html` Here is a small alert message. `,
+  defaultSlot: html`Here is a small alert message.`,
   size: 'small',
 };
 
--- a/src/components/base/outline-alert/outline-alert.stories.ts
+++ b/src/components/base/outline-alert/outline-alert.stories.ts
@@ -57,6 +57,8 @@ You can use a smaller alert with \`size\` of \`small\`.
 
 You can remove the icon with \`shouldShowIcon\` set to \`false\`.
 
+You can customize the header by adding a \`outline-alert--header\` slot.
+
 ## Accessibility
 
 If the alert has an interaction, you should indicate this with \`isInteractive\` set to \`true\`.
--- a/src/components/base/outline-alert/outline-alert.stories.ts
+++ b/src/components/base/outline-alert/outline-alert.stories.ts
@@ -106,6 +106,10 @@ Small.args = {
 
 export const Header = Template.bind({});
 Header.args = {
-  headerSlot: html`<span slot="outline-alert--header">Here is an alert with a custom header.</span>`,
+  headerSlot: html`
+    <span slot="outline-alert--header">
+      Here is an alert with a custom header.
+    </span>
+  `,
   defaultSlot: html` Here is a message. `,
 };
--- a/src/components/base/outline-alert/outline-alert.stories.ts
+++ b/src/components/base/outline-alert/outline-alert.stories.ts
@@ -94,8 +94,8 @@ const Template = ({
     <outline-alert
       statusType="${ifDefined(statusType)}"
       size="${ifDefined(size)}"
-      shouldShowIcon="${ifDefined(shouldShowIcon)}"
-      isInteractive="${ifDefined(isInteractive)}"
+      shouldShowIcon="${ifDefined(shouldShowIcon ? 'true' : undefined)}"
+      isInteractive="${ifDefined(isInteractive ? 'true' : undefined)}"
     >
       ${ifDefined(headerSlot)} ${ifDefined(defaultSlot)}
     </outline-alert>
--- a/src/components/base/outline-alert/outline-alert.stories.ts
+++ b/src/components/base/outline-alert/outline-alert.stories.ts
@@ -49,7 +49,7 @@ This component renders an alert.
   statusType="{{ statusType }}"
   size="{{ size }}"
 >
-  <outline-heading slot="outline-alert--header">{{ headerSlot }}</outline-heading>
+  <span slot="outline-alert--header">{{ headerSlot }}</span>
   {{ defaultSlot }}
 </outline-alert>
         `,
@@ -106,6 +106,6 @@ Small.args = {
 
 export const Header = Template.bind({});
 Header.args = {
-  headerSlot: html` Here is an alert with a custom header. `,
+  headerSlot: html`<span slot="outline-alert--header">Here is an alert with a custom header.</span>`,
   defaultSlot: html` Here is a message. `,
 };
--- a/src/components/base/outline-alert/outline-alert.stories.ts
+++ b/src/components/base/outline-alert/outline-alert.stories.ts
@@ -34,12 +34,24 @@ export default {
       options: alertSizes,
       control: { type: 'select' },
     },
+    shouldShowIcon: {
+      description: '(not yet implemented) Should we show the icon',
+      control: { type: 'boolean' },
+    },
   },
   parameters: {
     docs: {
       description: {
         component: `
 This component renders an alert.
+
+## Variation
+
+You can set the type of alert with \`statusType\`.
+
+You can use a smaller alert with \`size\` of \`small\`.
+
+You can remove the icon with \`shouldShowIcon\` set to \`false\`.
 `,
       },
       source: {
@@ -48,6 +60,7 @@ This component renders an alert.
 <outline-alert
   statusType="{{ statusType }}"
   size="{{ size }}"
+  shouldShowIcon="{{ shouldShowIcon }}"
 >
   <span slot="outline-alert--header">{{ headerSlot }}</span>
   {{ defaultSlot }}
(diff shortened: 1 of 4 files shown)


src/components/base/outline-alert/outline-alert.ts
```diff
@@ -0,0 +1,64 @@
...
+            `
+          : html``}
+        <div id="message">
+          <slot></slot>
+        </div>
+      </div>
+    `;
+  }
+
+  @property({ type: Boolean })
```
madhaze: The `@property`'s should be moved above the `render()`. Also can we add doc blocks for what these props control.

src/components/base/outline-alert/outline-alert.ts
```diff
@@ -0,0 +1,64 @@
...
+        ${this.shouldShowIcon === true
+          ? html`
+              <div id="icon">
+                <!--@todo include icon when we have that ready.-->
+              </div>
+            `
+          : html``}
+        ${this.size === 'large'
+          ? html`
+              <div id="header">
```
madhaze, line 67: Should this use the `<outline-header>`

madhaze:
We should also look at adding the ability to close alerts. If these might be used to display an alert dynamically after page load we should look into supporting aria-live polite

src/components/base/outline-alert/outline-alert.ts
damontgomery: Thanks for the question. I'm not sure.

In the storybook examples I left it as a slot, so we could pass in an `<outline-header>` if desired.

The trade-off there seems to be a little more work on the consumer of the component for the flexibility to define the header level / visual level.

I suppose we could always look to see if the slotted component was an `<outline-header>`, text, or something else and then only wrap text in the `<outline-header>`. What levels would we use then as defaults?

Anyone have ideas?

damontgomery:
> We should also look at adding the ability to close alerts. If these might be used to display an alert dynamically after page load we should look into supporting aria-live polite

Thanks, Chris!

I added this idea and the icon TODO to this story (#67) in case we want to implement the basic version first and then add enhancements.

src/components/base/outline-alert/outline-alert.ts
damontgomery: Thanks, @madhaze!

I appreciate the review and your perspective. It's incredibly helpful, because while I'm trying to create clear code, I don't have the perspective of someone else. :) I can only make guesses.

I've looked at the Lit examples (https://lit.dev/docs/) and see that the expectation is to put the properties first. I think that expectation makes sense and I got a little carried away with putting the `render()` method first.

I've also been trying to limit comments when the purpose of variables / properties / methods is clear, but I think the `why` of `isInteractive` is not so clear. What do you think of the others like `size`, `statusType`, `shouldShowIcon`?

Thanks again! Always appreciate the feedback.

src/components/base/outline-alert/outline-alert.ts
madhaze: Good point.  I saw the `this.statusType` and was thinking that was the text that was passed, but realize that is just a placeholder and we should be using content in the slot so that might make more sense.  One reason I guess would be to keep all headings the same style.

damontgomery:
Added an automated test. This should be ready for review. :)

src/components/base/outline-alert/outline-alert.ts
```diff
@@ -0,0 +1,78 @@
...
+  shouldShowIcon = true;
+
+  @property({ type: String })
+  size: AlertSize = 'large';
+
+  render(): TemplateResult {
+    // The `body` wrapper is used to avoid styles (like border) that are preventing us from styling `:host`.
+    return html`
+      <div id="body" role="${this.isInteractive ? 'alertdialog' : 'alert'}">
+        ${this.shouldShowIcon === true
```
mabry1985, line 58: @damontgomery This is looking pretty good! One piece I really would like to see changed are these two conditionals placed in their own template logic. It gets a little messy having these all in the render function as is

My suggestion is to make a couple template functions below the render function for each conditional:

```
  iconTemplate = (): TemplateResult | undefined => {
    if (this.shouldShowIcon) return

    return html`
      <div id="icon"></div>
    `
  }

  largeTemplate = (): TemplateResult | undefined => {
    if (this.size !== 'large') return

    return html`
      <div id="header">
        ${this.statusType}
        <slot name="outline-alert--header"></slot>
      </div>
    `
  }
```

then the render function can just call those and it makes it much easier to reason about and also gives the ability to overwrite these templates if someone wanted to extend this one.
```
  render(): TemplateResult {
    // The `body` wrapper is used to avoid styles (like border) that are preventing us from styling `:host`.
    return html`
      <div id="body" role="${this.isInteractive ? 'alertdialog' : 'alert'}">
        ${this.iconTemplate()}
        ${this.largeTemplate()}
        <div id="message">
          <slot></slot>
        </div>
      </div>
    `;
  }

```

I think we can also lose the comment on line 55, I don't think it's doing much for us.

src/components/base/outline-alert/outline-alert.ts
mabry1985: I agree with Chris on this one, I think this should be an `outline-heading` with the statusType rendered inside of it. I don't think there is much benefit to allowing a consumer to place their own element here. If someone wanted to do that they could extend this element and overwrite the `largeTemplate()` method I'd mentioned above.

src/components/base/outline-alert/outline-alert.stories.ts
```diff
@@ -0,0 +1,158 @@
...
+    },
+    isInteractive: {
+      description: 'Is there an interaction in the alert, such as a button.',
+      control: { type: 'boolean' },
+    },
+  },
+  parameters: {
+    docs: {
+      description: {
+        component: `
```
mabry1985, line 49: Could we clean up this formatting a bit? It gets a bit difficult to follow what's going on here

mabry1985 (changes requested):
The component overall looks good, I requested a few changes here that are mainly refactors and clean up.

src/components/base/outline-alert/outline-alert.stories.ts
damontgomery: Thanks for taking a look. Could you give some examples of what's difficult?

The overall formatting is forced by Prettier and this template literal I put the markdown aligned to the left because I find that easier to read. It's like a HEREDOC / NOWDOC. Did you want these indented to the `component` property?

mabry1985: Yeah sorry I wasn't more explicit, I meant just indenting this and the `code` at line 70. If this is a prettier issue maybe we can just throw a prettier ignore comment on these trouble areas

src/components/base/outline-alert/outline-alert.ts
```diff
@@ -0,0 +1,78 @@
...
+  statusType: AlertStatusType = 'information';
+
+  /**
+   * This is important context for screen readers.
+   */
+  @property({ type: Boolean })
+  isInteractive = false;
+
+  @property({ type: Boolean })
+  shouldShowIcon = true;
```
damontgomery, line 49: This probably needs to be `shouldHideIcon` because of the way attributes work in the DOM. The presence of an attribute means `true`, so there is no way to set a false value if the default is `true`. :(

So, all boolean attributes probably need to default to false.

src/components/base/outline-alert/outline-alert.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  @property({ type: String })
+  size: AlertSize = 'large';
+
+  render(): TemplateResult {
+    // The `body` wrapper is used to avoid styles (like border) that are preventing us from styling `:host`.
+    return html`
+      <div id="body" role="${this.isInteractive ? 'alertdialog' : 'alert'}">
```
LeeMellon, line 57: I think this is going to require some more 'plumbing'. 
Looks like you're going to need for 'alertdialog': 
 - aria-label/ledby 
 - focusable control/focused close button. 
 - trapped/wrapping tab order.
 https://developer.mozilla.org/enUS/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alertdialog_role

I see the SB example, but we can't always be sure that the consumer will pass the requirements.

src/components/base/outline-alert/outline-alert.ts
```diff
@@ -0,0 +1,78 @@
...
+export const alertStatusTypes = [
+  'information',
+  'warning',
+  'error',
+  'success',
+] as const;
+export type AlertStatusType = typeof alertStatusTypes[number];
+
+// This can be useful for testing.
+export interface OutlineAlertInterface extends HTMLElement {
```
LeeMellon, line 18: I like this a lot and think it should become a standard.

src/components/base/outline-alert/outline-alert.ts
```diff
@@ -0,0 +1,78 @@
...
+  @property({ type: String })
+  statusType: AlertStatusType = 'information';
+
+  /**
+   * This is important context for screen readers.
+   */
+  @property({ type: Boolean })
+  isInteractive = false;
+
+  @property({ type: Boolean })
```
LeeMellon, line 48: Even if they're simple/obvious let's keep the standard of docs for every/state/method.

src/components/base/outline-alert/outline-alert.stories.ts
```diff
@@ -0,0 +1,158 @@
...
+      size="${ifDefined(size)}"
+      ?shouldShowIcon="${shouldShowIcon}"
+      ?isInteractive="${isInteractive}"
+    >
+      ${ifDefined(headerSlot)} ${ifDefined(defaultSlot)}
+    </outline-alert>
+  `;
+};
+
+export const Information = Template.bind({});
```
LeeMellon, line 105: I don't think we need a SB example for every possible iteration/variant.

src/components/base/outline-alert/outline-alert.ts
```diff
@@ -0,0 +1,78 @@
+import { html, TemplateResult, CSSResultGroup } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
+import componentStyles from './outline-alert.css.lit';
+import { OutlineElement } from '../outline-element/outline-element';
+
+export const alertSizes = ['small', 'large'] as const;
```
LeeMellon, line 6: I think this is over complicating things. I think you want to replace these two lines with `export type AlertSize = 'small' | 'large'`

damontgomery: Thanks for the review @LeeMellon. Going to look into all this in more detail when I get a chance.

The attempt here was to provide an array of values used by Storybook to populate the select box and provide the type in this file for TS checking.

Is there a way to do both of those in one place? This pattern I found somewhere, but was new to me. :) The best I could find to avoid copy & paste issues.

src/components/base/outline-alert/outline-alert.ts
damontgomery: Thanks for the links! Maybe we can remove this feature for now and add a todo.

src/components/base/outline-alert/outline-alert.ts
mabry1985: No this is good, he's using [const assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) here. I just figured out how to do this myself this week, it allows us to create a type and capture it in a variable in one go.

src/components/base/outline-alert/outline-alert.ts
mabry1985: Yeah I agree, I've been doing this and reflecting that comment in the story controls. I'd like to look into creating a method for using those JSDocs to populate the storybook control info for us. Stencil does something similar.

I think for Storybook, especially, it's important to be explicit with the comments because there are QA, design, clients, etc. who are going to be seeing these pages. Its more user friendly and accessible is my argument. 

A Storybook style guide is something I'd love to get nailed down soon, I'm just trying to figure out the best way to write them still too. Would be a good discussion for guild.

src/components/base/outline-alert/outline-alert.ts
mabry1985: Interesting 🤯

src/components/base/outline-alert/outline-alert.stories.ts
mabry1985: I think it's beneficial to write stories for every visual representation of a component like we do with the button component, especially as we implement VRT. It would be good to detect if something you did broke a layout of a component in a specific state.

src/components/base/outline-alert/outline-alert.stories.ts
```diff
@@ -0,0 +1,158 @@
...
+    },
+    statusType: {
+      description:
+        'The status type of the alert. Such as `information` or `warning`.',
+      options: alertStatusTypes,
+      control: { type: 'select' },
+    },
+    size: {
+      description: 'The size of the alert.',
+      options: alertSizes,
```
mabry1985, line 34: @LeeMellon here is where the const assertion mentioned below comes in handy, now if you need to add a new alert size or modify an existing one, you just add it in the one file and you're good to go.

src/components/base/outline-alert/outline-alert.stories.ts
LeeMellon: I'm down to have this discussion in a larger setting, but for me SB serves t purposes. 1: To show what the component can do. 2: To show how to use the component. 
I think 1 is taken care of by the controls we take the time to set up. So the number of variants in a story should be dictated by the different ways the component can be used. EX: the alertdialog option here.

src/components/base/outline-alert/outline-alert.ts
LeeMellon: I think you could change `isInteractive` to something like `alertDialogue: string` and if it's there pass that to an aria-label attrib, and then set the role to "alertdialog". Just a thought to keep the functionality. Also we may just need to have a focusable close button here either way.

src/components/base/outline-alert/outline-alert.stories.ts
damontgomery: I think when we tried this with indenting and a template literal, the Markdown didn't work. :( I think Markdown needs the header / etc syntax at the start of the line, `#`.

src/components/base/outline-alert/outline-alert.stories.ts
damontgomery: @LeeMellon , I think that's an interesting point.

This is the first time I've used storybook, so the patterns are new to me. I can see the point of view that a smaller number of stories may be easier for an implementer since it focuses their attention.

If it's helpful, this is what I've used Storybook for so far.

I write the storybook page first with examples of the variations that I understand need to be represented. In the alert, this means one of each status type. There is some level of feeling things out here. On a current project, we have an icon that can show 1800 icons and I don't have 1800 stories. So, I think 4 statuses makes sense to me as stories, but 1800 doesn't. I made another component to show all the colors / icons and those have their own component pages. In development, I was settings CSS classes per status type, but there were no CSS changes for those icons. Meaning, it was possible for me to break a single status type, but not a single icon.

If there is JS interaction, I may write the Jest tests at this point too.

Then, I start implementing the component itself. At this time, I'm watching those stories to see the components change and gradually get the features that I expect. Most of the time I'm looking at the `docs` pane, or for interaction / screen reader the `iframe` example.

If I forgot an example, I'll add more stories to capture those use cases.

... And I don't really use the controls much.

I agree with @mabry1985, that for visual regression testing, some of the additional variations may be helpful since the tools will default to one screenshot of each.

Might be a good team discussion or one for Github discussions.

src/components/base/outline-alert/outline-alert.ts
damontgomery: Thanks for the explanation, @mabry1985! And the link to where it's used.

LeeMellon: Thanks both of you. Using this regularly now.

src/components/base/outline-alert/outline-alert.ts
LeeMellon: After a while on a project with the enhanced editor I think the slot is a better idea.

LeeMellon (approved):

himerus:
I know this one has been sitting a while, but I'm about ready to start the "mass merge". However, I want to hold on this one and update the `outline-icon` in https://github.com/phase2/outline/pull/118. I figure there will need to be a minor adjustment to this MR then.

PR #63, opened 2021-07-18
add testing and refactor stories to work with it

mabry1985:
## Testing steps

Testing
- Remove `node_modules`
- Run `yarn install`
- Run `yarn test`
- ...?

Storybook
- Run `yarn start`
- Confirm stories are still rendering as expected. (you don't get an error when viewing components).

### Add a known failing test (optional)

Known failing test can be added in `src/components/base/outline-button/test/outline-button.test.ts`
```
describe('failing test', () => {
  it('expects to fail', () => {
    assert.equal(1, 2, "1 is not 2")
  });
});
```

Then run `yarn test` again and see the failure notice.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "rollup.bundle": "rollup -c -m",
     "rollup.watch": "yarn rollup -w",
     "rollup": "yarn rollup.bundle",
-    "test": "tsc && wtr"
+    "test": "yarn build && wtr"
   },
   "keywords": [
     "web-components",
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "@babel/plugin-proposal-decorators": "^7.14.2",
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
-    "@open-wc/testing": "^2.5.33",
+    "@open-wc/testing": "^3.0.0-next.2",
     "@rollup/plugin-multi-entry": "^4.0.0",
     "@rollup/plugin-replace": "^2.4.2",
     "@storybook/addon-a11y": "^6.3.4",
--- a/package.json
+++ b/package.json
@@ -59,7 +59,8 @@
     "reboot": "yarn clean && yarn reset && yarn install && yarn start",
     "rollup.bundle": "rollup -c -m",
     "rollup.watch": "yarn rollup -w",
-    "rollup": "yarn rollup.bundle"
+    "rollup": "yarn rollup.bundle",
+    "test": "tsc && wtr"
   },
   "keywords": [
     "web-components",
@@ -74,6 +75,7 @@
     "@babel/plugin-proposal-decorators": "^7.14.2",
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
+    "@open-wc/testing": "^2.5.33",
     "@rollup/plugin-multi-entry": "^4.0.0",
     "@rollup/plugin-replace": "^2.4.2",
     "@storybook/addon-a11y": "^6.3.4",
@@ -92,6 +94,9 @@
     "@typescript-eslint/parser": "^2.27.0",
     "@web/dev-server": "0.0.29",
     "@web/dev-server-legacy": "^0.1.4",
+    "@web/test-runner": "^0.13.15",
+    "@web/test-runner-browserstack": "^0.4.2",
+    "@web/test-runner-playwright": "^0.8.6",
     "@webcomponents/webcomponentsjs": "^2.5.0",
     "babel-loader": "^8.2.2",
     "concurrently": "^6.0.2",
--- a/src/components/base/outline-card/outline-card.stories.ts
+++ b/src/components/base/outline-card/outline-card.stories.ts
@@ -1,4 +1,4 @@
-import { html, TemplateResult } from 'lit-html';
+import { html, TemplateResult } from 'lit';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { argTypeHidden } from '../outline-element/utils/utils';
 import sampleImage1 from '../../../.storybook/static/media/color/Codeworkbench_color.jpg';
--- a/src/components/base/outline-card/outline-card.stories.ts
+++ b/src/components/base/outline-card/outline-card.stories.ts
@@ -1,4 +1,4 @@
-import { html, TemplateResult } from 'lit';
+import { html, TemplateResult } from 'lit-html';
 import { ifDefined } from 'lit/directives/if-defined.js';
 import { argTypeHidden } from '../outline-element/utils/utils';
 import sampleImage1 from '../../../.storybook/static/media/color/Codeworkbench_color.jpg';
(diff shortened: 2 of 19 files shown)


package.json
```diff
@@ -59,7 +59,8 @@
     "reboot": "yarn clean && yarn reset && yarn install && yarn start",
     "rollup.bundle": "rollup -c -m",
     "rollup.watch": "yarn rollup -w",
-    "rollup": "yarn rollup.bundle"
+    "rollup": "yarn rollup.bundle",
```
damontgomery, line 62: possible dumb question:

Are we using webpack for Storybook and rollup for the production package?

damontgomery: Maybe this should go in the Q&A on this repo...

src/components/base/outline-card/outline-card.stories.ts
```diff
@@ -96,7 +96,7 @@ const Template = ({
   html`
     <outline-container>
       <outline-card
-        hasshadow="${ifDefined(hasShadow)}"
+        ?hasshadow=${ifDefined(hasShadow)}
```
damontgomery, line 99: Thanks for this. In reviewing the PR I double checked and realized this worked for the alert code. :)

src/components/base/outline-link/outline-link.stories.ts
```diff
@@ -1,4 +1,4 @@
-import { html, HTMLTemplateResult } from 'lit';
+import { html, TemplateResult } from 'lit-html';
```
damontgomery: Is there a difference between the `html` in `lit` and `lit-html` and a reason for picking one or the other?

I noticed in one of the components that `lit-html` doesn't have `CSSResultGroup`, but `lit` does. `outline-button` uses this as an example.

src/components/base/outline-card/outline-card.stories.ts
mabry1985: Oh nice, yeah I wanted to talk about that during the dev huddle today. I cleared up some confusion I had around the data binding.

package.json
mabry1985: @himerus would have an answer for you on this, I'm not 100% sure.

himerus: Yes, Webpack currently for Storybook until the faster bundlers like vite/snowpack are more stable. 
Storybook directly can consume the .ts stuff so no compiling is needed to see updates during development. 
Rollup for the production (bundled) build and standard tsc for the build that has the components broken out.

src/components/base/outline-link/outline-link.stories.ts
himerus: I think we are exclusively supposed to be using the `lit` package only now, however, this seems like a work around for this issue with what was going on with stories no longer rendering. @mabry1985 did you have a link to an issue in Lit (or openwc/testing) related to this?

mabry1985: So previously the stories were breaking when we installed the `@open-web/testing` dependency and pulling `html` straight from `lit-html` instead of `lit` fixed the issue. Which is strange because if you look at the `html` import from `lit` it seems to be just passing through `html` from `lit-html`

We would still exclusively use the imports from `lit` when building the components themselves though.

mabry1985: @himerus I'm currently in discussion with someone on the Lit slack about seeing if we can sort this out I'll open an issue if that doesn't get matters resolved. I think the workaround is okay for now though.

package.json
```diff
@@ -59,7 +59,8 @@
...
+    "test": "tsc && wtr"
```
himerus: My question is do we actually have to run build for this? Or can `wtr` operate on the typescript itself?

I think this should change to `yarn build && wtr` though. Just making sure we run the normal build processes.

damontgomery: TIL you can comment the `package.json` with `//` keys.

https://stackoverflow.com/a/14221781

Might be helpful for short named things like the TypeScript compiler and Web Test Runner?

`"//": "Run the TypeScript compiler and the Web Test Runner.",`

damontgomery (commented):
Nice! I was able to get this to run as expected. But, please let me know if I can help by reviewing something else too.

## Questions
- The `yarn.lock` file is changed after I run `yarn install` can we review that?

## Testing steps

- deleted `node_modules`
- Installed packages with `yarn install`
- run `yarn test`
- Confirm everything passes
- Add a known failing test (see below)
- run `yarn test`
- Confirm one test fails

Known failing test added in `/Users/dan/Sites/outline/src/components/base/outline-button/test/outline-button.test.ts`
```
describe('failing test', () => {
  it('expects to fail', () => {
    assert.equal(1,2,"1 is not 2")
  })
})
```

## Results

```
: yarn test
yarn run v1.22.10
$ tsc && wtr

Chromium: |██████████████████████████████| 9/9 test files | 25 passed, 0 failed
Firefox:  |██████████████████████████████| 9/9 test files | 25 passed, 0 failed
Webkit:   |██████████████████████████████| 9/9 test files | 25 passed, 0 failed

Finished running tests in 13.1s, all tests passed! 🎉

✨  Done in 22.91s.
: yarn test
yarn run v1.22.10
$ tsc && wtr

dist/components/base/outline-button/test/outline-button.test.js:

 ❌ failing test > expects to fail
      AssertionError: 1 is not 2: expected 1 to equal 2
      + expected - actual

      -1
      +2

      at o.<anonymous> (src/components/base/outline-button/test/outline-button.test.ts:6:11)

Chromium: |██████████████████████████████| 9/9 test files | 25 passed, 1 failed
Firefox:  |██████████████████████████████| 9/9 test files | 25 passed, 1 failed
Webkit:   |██████████████████████████████| 9/9 test files | 25 passed, 1 failed

Finished running tests in 8.3s with 3 failed tests.

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
:
```

package.json
mabry1985: I get yelled at when I try to do this

Screen Shot 2021-07-19 at 10 33 14 AM

damontgomery: You can't add a comment at the end of a line, the key needs to be `//` as if it were another dependency / command / property on the object. It is confusing. :(

It looks like it's a node work around for the limitations of JSON.

mabry1985: oh interesting

damontgomery: If I haven't run storybook (or compiled the stylesheets), then the TypeScript compiler complains about the references to non-existing files.

```
src/components/base/outline-accordion/outline-accordion.ts:3:29 - error TS2307: Cannot find module './outline-accordion.css.lit' or its corresponding type declarations.

3 import componentStyles from './outline-accordion.css.lit';
```

Should we add `generate.component-css` to this task?

damontgomery (commented):
Overall this looks good to me!

I left a comment about the compiled styles for the yarn command.

Still also getting changes in `yarn.lock`, but I don't know what to make of that.

Storybook works well and shows the elements. It also doesn't have issues with empty attributes as it did at one point. Those would show up as `Symbol(lit-nothing)`.

package.json
himerus: running `yarn build` instead of `tsc` in this `test` script will solve the above.

himerus: @mabry1985 If we change the above `yarn test` command as described, I think we may be good to go here.

himerus (changes requested):

himerus (approved):

himerus:
After the last commit, including a new `yarn.lock`, this seems to install cleanly with a `yarn reboot`.

PR #66, opened 2021-07-20
Update button component

LeeMellon:
##Minor update to button 
- Handles 'aria-disabled' attribute for  `<outline-button>` and sub/shadow root elements. 
- Adds onClick and onKeyUp props. 
- Adds css styles to prevent button clicks on disabled. 

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/66"></a>


src/components/base/outline-button/outline-button.stories.ts
```diff
@@ -85,6 +89,14 @@ export const PrimaryButton = Template.bind({});
 PrimaryButton.args = {
   defaultSlot: 'Primary button',
   variant: 'primary',
+  onClick: (e: Event) => {
+    e.preventDefault(), console.log('onClick called!!');
```
mabry1985: console.log statement here

src/components/base/outline-button/outline-button.ts
```diff
@@ -49,19 +49,55 @@ export class OutlineButton extends OutlineElement {
...
+   */
+  @property() onClick: () => void;
+
+  /**
+   * A keyUp handler to be passed to the @keyUp attribute
+   */
+  @property() onKeyUp: () => void;
+
+  /**
+   * Sets role attribute to "button", for accessibility purposes.
```
damontgomery: This comment is helpful!

src/components/base/outline-button/outline-button.ts
```diff
@@ -49,19 +49,55 @@ export class OutlineButton extends OutlineElement {
...
+   */
+  @property({ type: String, reflect: true }) role = 'button';
+
+  /**
+   * Sets tabindex attribute to default "0", for accessibility purposes.
```
damontgomery: Another helpful comment!

src/components/base/outline-button/outline-button.ts
```diff
@@ -49,19 +55,105 @@ export class OutlineButton extends OutlineElement {
...
+
+  /**
+   * Sets the position of the tooltipInfo/disabledInfo display on hover.
+   */
+  @property({ type: String }) position: TooltipPosition = 'bottom';
+
+  /**
+   * Used to provied a unique ID to panel content <div> for accessibility purposes.
+   */
+  seed = Math.floor(Math.random() * 10000);
```
damontgomery: Could this be part of the logic in the `outline-tooltip` component? Are there cases when it's helpful for the consumer to provide the ID or could be generate random strings in all situations?

damontgomery (commented):
Thanks again for all the work here.

I tested by reviewing the button code and the button stories in Storybook.

## Notes
- The comments in the button component do a good job of explaining `why` the properties / etc exist which is helpful context. :)

## Questions
- Is there a way to use the keyboard to select the buttons? I can't get it to work. Visually it doesn't show and looking at `document.activeElement` in the console doesn't show the button. Focus with JS does work, `$('outline-button').focus()` in the console.
- (improvement) Could we have examples for disabled buttons / tooltips in the stories?

Thanks again.

src/components/base/outline-button/outline-button.css
```diff
@@ -1,5 +1,11 @@
+:host {
+  /* flex and border-radius constrain the actual boundaries of the element,
+   preventing literal edge cases where a user can click on the edge or corner of the button and defeate states like disabled */
+  @apply flex rounded-xl;
+}
+
 .btn {
-  @apply px-12 py-4 text-lg leading-tight rounded-xl cursor-pointer transition-colors duration-300 font-body no-underline inline-flex content-center;
```
mabry1985, line 2: Can we add the padding back here? I thought `py-3 px-4` looked pretty good.

mabry1985 (changes requested):
@LeeMellon I think this button is looking great and I used it mostly as is on SMC. Can we pull the tooltip from this though and work on that separately in [this](https://github.com/phase2/outline/pull/74) ticket? 

We can extend this base button and implement tooltip functionality there.

src/components/base/outline-button/outline-button.css
LeeMellon: It's still there for large buttons.

LeeMellon:
Tooltip is here because it was part of the requested functionality. So let's get tool-tip up to snuff first and then merge this in.

src/components/base/outline-button/outline-button.ts
LeeMellon: tooltip requires an id so that the target can be associated with the tooltip. It seemed best for the id to be user/consumer generated incase they wanted/needed to specify. In this scenario I hardcoded it as it seemed like stable behavior. But I'm open to removing the id prop entirely from the tooltip and letting it be autogenerated within the tooltip itself.

mabry1985 (approved):
Looks good!

mabry1985:
@shaal @himerus if one of you have a chance to take a look at this, I think it's good to go.

src/components/base/outline-button/outline-button.stories.ts
```diff
@@ -110,17 +134,30 @@ IconLink.args = {
 export const PrimaryButton = Template.bind({});
 PrimaryButton.args = {
   defaultSlot: 'Primary button',
+  size: 'medium',
   variant: 'primary',
+  onClick: (e: Event) => {
+    e.preventDefault();
+    console.log('onClick called!!');
```
shaal:
```suggestion
```

src/components/base/outline-button/outline-button.stories.ts
```diff
@@ -110,17 +134,30 @@ IconLink.args = {
...
+  },
+  onKeyUp: (e: KeyboardEvent) => {
+    e.preventDefault();
+    if (e.key === ' ') {
+      console.log(e);
```
shaal:
```suggestion
      console.error(e);
```

src/components/base/outline-button/outline-button.css
```diff
@@ -13,23 +19,21 @@
...
   &:hover,
   &:focus {
     @apply bg-neutral-black text-neutral-white;
   }
-
-  &:disabled {
-    @apply cursor-not-allowed text-gray-500 bg-gray-300;
-  }
+}
+.btn.secondary[aria-disabled='true'] {
```
madhaze: Should we move all the disable styles to one place and just separate the selectors with comma. Instead of repeating the same apply three times. Or does just ‘.btn[aria-disabled=“true”]’ work like the .small classes?

mabry1985:
@LeeMellon is this PR is an outdated version of https://github.com/phase2/outline/pull/93 ?

LeeMellon:
@mabry1985 No this is just updating the button on its one. I wasn't sure if the built-in tooltip was going to fly, or if it should be a sub version of button.

madhaze (approved):

PR #71, opened 2021-07-20
docs: Example html consumer.

damontgomery:
This example is intended to give some idea of how people can create consumers.

Looking for feedback as always. :)

## Testing notes

- Review the readme included in the folder
- Build outline
- Start a webserver
- Review the page
- Review the HTML source

## Example

html_example

## Questions

The page requested all those files like `package.json`, so I symlinked to them like `ln -s ../../package.json package.json`. It seems to generate without them too, so :shrug:.

### DIFF
--- a/example-consumers/html/README.md
+++ b/example-consumers/html/README.md
@@ -1,0 +1,24 @@
+# Example site to illustrate loading the web components from Outline
+
+## Build the project
+
+```
+yarn install
+yarn build
+```
+
+## Install a simple web server
+
+`npm install --global http-server`
+
+## Run the simple web server
+
+From inside the this folder
+
+`npx http-server`
+
+## View the example
+
+Visit [http://127.0.0.1:8080/](http://127.0.0.1:8080/)
+
+_Your `http-server` output may use a different port._
--- a/example-consumers/html/index.html
+++ b/example-consumers/html/index.html
@@ -1,0 +1,36 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Outline demo</title>
+  <style>
+    :root { font-family: sans-serif; }
+  </style>
+  <link rel="stylesheet" href="outline.theme.css">
+  <script type="module" src="outline.js"></script>
+</head>
+<body>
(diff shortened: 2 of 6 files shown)


damontgomery:
Closing so we can update `yarn serve` and `/dev` examples.

damontgomery:
I created a test POC branch here: https://github.com/phase2/outline/blob/poc/mdx-stories/src/components/base/outline-alert/test/outline-alert.test.ts

## Open questions

- Is there a better way to load default CSS?
- Is there a better way to load the necessary JS for each component?
- Is there a better way to handle render time for components?

### Loading CSS

The interactive test runner `yarn wtr --watch` provides a nice way to open the browser and see what's happening during tests. For this to be more useful, it would be nice to load the default CSS.

The `outline.theme.css` isn't loaded by default.

I got around this with this
```javascript
before(() => {
    const defaultCSS = document.createElement('link');
    defaultCSS.setAttribute('rel', 'stylesheet');
    defaultCSS.setAttribute('src', '/outline.theme.css');

    document.head.append(defaultCSS);
  })
```

### Loading component JS

If we don't use the `Outline*` object from the component, the test won't load it by default. If we import the module, but don't use it, TypeScript complains...

I got around this with adding a meaningless test.

```javascript
it('is defined', () => {
    const alertElement = document.createElement('outline-alert');
    alertElement.setAttribute('statusType', 'error');

    // If we don't use OutlineAlert at least once, it's JS will never be loaded and the component won't render. :(
    assert.instanceOf(alertElement, OutlineAlert);

    document.body.append(alertElement);

    const alertElementInDOM = document.querySelectorAll('outline-alert');
    expect(alertElementInDOM).to.exist;
  });
```

### Waiting for the render of components

If we don't wait for components to render, their properties / slots / etc won't be accurately tested.

I got around this by waiting a small amount of time.

```javascript
setTimeout(() => {
  expect(messageElement.assignedSlot, 'Message is not slotted.').is.not.null;
 }, 10);
```

I tried to see if Lit elements emit any events that we could respond to, but I couldn't find anything about that.

I tried `wait-for-expect`, but had issues. See https://github.com/TheBrainFamily/wait-for-expect/issues/31.

Could we use CypressJS which automatically retries to avoid these types of brittle tests?

## Other feedback

I think this test works OK. It doesn't check anything about the implementation so far other than that the slots are slotted. Not sure how to check if text shows up on the screen given that all the text checkers in `chai-dom` will show true since the content is actually in the light DOM even if it's not visible.

PR #74, opened 2021-07-22
feat(tooltip): add outline-tooltip

LeeMellon:
## Outline-Tooltip
Cursor_and_Atoms___OutlineTooltip_-_Inline_Tooltip_⋅_Storybook_and_iangoodrich_elfayoumi____dev_internal_outline

### Adds lightly styled `<outline-tooltip>` component.
- Works as a wrapper around content dev wants to provide with a tooltip.
- Can be passed tip content via "tip" property or "tip-info" slot.
- Can be cardinally positioned. 

## Note:
   Is also included in PR #66 as proposed button update includes built-in tooltip functionality. 

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/74"></a>


src/components/base/outline-tooltip/outline-tooltip.stories.ts
```diff
@@ -0,0 +1,87 @@
...
+        options: ['top', 'bottom', 'left', 'right'],
+      },
+    },
+  },
+};
+
+const InlineTemplate = ({ id, tip, position }): TemplateResult =>
+  html`
+    <div style=${`max-width: 50%`}>
+      <h2 style=${`font-size: 30px; margin-bottom: 1rem`}>
```
damontgomery: Could we use `outline-heading` and `outline-grid` or `outline-column` for this example?

src/components/base/outline-tooltip/outline-tooltip.stories.ts
```diff
@@ -0,0 +1,87 @@
...
+const FormTemplate = ({ id, position }): TemplateResult =>
+  html`
+    <div>
+      <h2 style=${`font-size: 30px; margin-bottom: 1rem`}>
+        Simulated Disabled Form Button.
+      </h2>
+      <div>
+        <form>
+          <outline-tooltip id="${id}" position="${position}">
+            <span slot="tip-info"> Enter a valid email adress here </span>
```
damontgomery: (very minor) spelling on `address`.

src/components/base/outline-tooltip/outline-tooltip.ts
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   *  REQUIRED: An ID string that associated components will need for
+   *  accessibility/screen reader compliance.
+   */
+  @property({ type: String }) id!: string;
+
+  /**
+   * Identifies the component as a tooltip for accessibility technologies.
+   */
+  @property({ type: String, reflect: true }) role = 'tooltip';
```
damontgomery: Are we expecting the consumer to change this? If not, should we make this part of the template directly?

Is this used? I don't see it in the template.

src/components/base/outline-tooltip/outline-tooltip.stories.ts
```diff
@@ -0,0 +1,87 @@
+import { html, TemplateResult } from 'lit';
+import { ifDefined } from 'lit/directives/if-defined';
+import './outline-tooltip';
+import '../outline-button/outline-button';
+export default {
+  title: 'Atoms/OutlineTooltip',
```
damontgomery: Should we use `Tooltip` to align with the other atoms like `Link` on the Storybook sidebar?

src/components/base/outline-tooltip/outline-tooltip.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * An optional prop to pass the tip info. Can also be passed via the
+   * "tip-info" slot.
+   */
+  @property() tip: string | HTMLElement | void;
```
damontgomery: I haven't seen `void` used like this before, only as a function return type. Could this be `null`?

damontgomery (commented):
Works well overall. Thanks, @LeeMellon.

I tested this by reviewing the code and loading the component in Storybook.

## Notes
- Thanks for the testing notes in Github!
- The examples are helpful since they contains testing steps. :)
- Controls descriptions are also helpful and clear, particularly about the slot / attribute for the Tip info.
- Component comments are helpful too!

## Questions
- (minor) Could the tooltip show up more quickly? It took longer than expected for me.
- (improvement) Should we account for placement when position puts the tooltip off the screen?

I left a few comments / questions in the codebase too.

madhaze:
I'm not sure these are accessible. We should look at https://inclusive-components.design/tooltips-toggletips/

LeeMellon:
In what way exactly @madhaze ?

src/components/base/outline-tooltip/outline-tooltip.ts
mabry1985: I think this is a good point. Unless we are intending for this to be changed let's just plug it into the template directly.

src/components/base/outline-tooltip/outline-tooltip.stories.ts
mabry1985: This makes sense to me to omit `Outline` from the title here.

src/components/base/outline-tooltip/outline-tooltip.ts
mabry1985: I agree that `void` may not be the proper typing here, but I would also argue that it shouldn't be `null` but rather `undefined` because thats what would print in a console.log if this wasn't passed an argument.

src/components/base/outline-tooltip/outline-tooltip.css
```diff
@@ -0,0 +1,41 @@
...
+  @apply inline-flex flex-col;
+}
+
+.tooltip {
+  @apply flex relative justify-center;
+}
+
+.tooltip-tip {
+  @apply absolute inline-flex p-2 opacity-0 text-center;
+  visibility: hidden;
```
mabry1985, line 15: we can add `hidden` to the `@apply` directive to accomplish the `visibility: hidden` property here

src/components/base/outline-tooltip/outline-tooltip.css
```diff
@@ -0,0 +1,41 @@
...
+}
+
+.position-left {
+  @apply right-full mr-2;
+}
+
+.tooltip:hover > .tooltip-tip,
+.tooltip:focus > .tooltip-tip {
+  @apply opacity-100;
+  visibility: visible;
```
mabry1985, line 42: As above we can use `visible` in the `@apply` directive to handle the `visibility` property here

mabry1985 (changes requested):
I think overall this is looking pretty good. I added a couple minor changes that I think could be addressed. 

I also want to see what comes of Ofer's testing of the accessibility before this is merged. 

Also, let's write a separate ticket for styling this component, I can dig around and see what others are doing there and tackle that  once this is approved and merged.

src/components/base/outline-tooltip/outline-tooltip.ts
LeeMellon: All of the above are outdate comments that have been addressed. I should have collapsed them. My bad.

src/components/base/outline-tooltip/outline-tooltip.css
LeeMellon: No. They work differently. Hidden = display: 'none'; Which is great but it means you can't use transitions. Also I'm using the visibility css attribute to control the cursor events, so that you won't know the tip is there and can't trigger it or anything in it until the visibility is set to visible.

src/components/base/outline-tooltip/outline-tooltip.css
LeeMellon: ^^

src/components/base/outline-tooltip/outline-tooltip.css
mabry1985: Gotcha! Thanks for clearing that up for me

src/components/base/outline-tooltip/outline-tooltip.css
mabry1985: noted, thanks for the clarification

mabry1985 (changes requested):
Thanks for clearing up my confusion on the visibility properties. I think if we can pull in the accessibility additions @shaal suggested I'd be happy to call this ready to go.

himerus (changes requested):
I'm going to assume again, this could use some slight alterations and bring brought up to date. @shaal I would like you to review this for a11y, and determine if instead there's a library we should be using or provide alternatives if needed.

PR #75, opened 2021-07-23
Docs: Basic html consumer.

damontgomery:
Update the `yarn serve` / `yarn wds` to work correctly and update the example HTML file / readme to reflect current state.

## To test

- View the readme. ;)
- `yarn install`
- `yarn serve`

- visit [http://localhost:8000/dev](http://localhost:8000/dev)

It seems like this issue is also being affected by #79. As a symptom, some of the web components are not rendered, such as the `my-element` component. You can see that there is no shadow DOM when you look at those elements and they don't appear as they would in Storybook.

## Screenshot

outline_example

## Notes

The `yarn serve` command was giving issues and saying we needed to either convert the `web-test-runner.config.*` file to use `exports`, rename to `mjs`, or add `module` and the `type` in `package.json`.

The rename seemed the least impacting, so I went with that.

## References

Based on feedback from #71

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/75"></a>


himerus (changes requested):
I like this overall, however I see the following. 
* We need to be excluding the `examples` directory from being part of the production build.
* This means those would be unavailable on this HTML consumer that utilizes the built output.
* I think keep it simple and short, and stick to just a couple of examples of link and button or something. 
* I would also pull in `next`, and add the examples directory (as an exclusion) to the rollup config as part of this adjustment.

damontgomery:
> I like this overall, however I see the following.
> 
> * We need to be excluding the `examples` directory from being part of the production build.
> * This means those would be unavailable on this HTML consumer that utilizes the built output.
> * I think keep it simple and short, and stick to just a couple of examples of link and button or something.
> * I would also pull in `next`, and add the examples directory (as an exclusion) to the rollup config as part of this adjustment.

Thanks for the review @himerus!

I've made the changes you mentioned and I think this is ready for another review.

mabry1985:
@damontgomery I pulled this down ran `yarn install` and `yarn serve` and it doesn't appear to be rendering correctly for me
Screen Shot 2021-07-29 at 6 10 10 PM

damontgomery:
> @damontgomery I pulled this down ran `yarn install` and `yarn serve` and it doesn't appear to be rendering correctly for me
> Screen Shot 2021-07-29 at 6 10 10 PM

Thanks for looking @mabry1985!

I tried it again and didn't have the issue you were having. `yarn serve` should run `yarn rollup` and then `yarn wds`. Maybe try clearing `dist` first and waiting for the Rollup output to finish? Are you getting any network errors / other errors in the inspector?

rollup_output

mabry1985:
@damontgomery hmmm interesting, thanks for the info, I'll give it another shot this afternoon.

LeeMellon:
Works well locally.

mabry1985 (approved):
This is great, thanks @damontgomery 

@himerus if you're happy with the changes requested I think we should pull this in.

himerus (approved):

himerus:
I'm having odd issues as well, even after clearing `dist` and `node_modules`. Going to chalk it up to my environment, and if this isn't working properly for some reason, we can adjust.

Issue #79, opened 2021-07-23
Package from Rollup is missing some of the components.


damontgomery:
As-is, if we get to a certain size (I added outline-alert and outline-modal), Rollup wants to start chunking stuff and it seems to just toss away some random parts of the code or give an error.

I could test this by searching through the `dist/outline.js` file and couldn't find strings like `outline-alert`, but could find `outline-link`.

I think it has something to do with `plugin-multi-entry`. https://www.npmjs.com/package/@rollup/plugin-multi-entry

The closest I got was something like,

```json
input: {
    include: [
      'src/components/**/*.ts',
    ],
    exclude: [
      'src/components/**/*.stories.ts',
      'src/components/**/*.css.lit.ts',
    ],
  },
  output: [
    {
      // Modern bundle.
      dir: 'dist',
      format: 'esm',
      entryFileName: 'outline.js',
    },
  ],
  manualChunks: () => 'outline.js',
```

You can run this with `yarn rollup.watch`.

But this uses the default name of `multi-entry.js` even though I've tried setting `entryFileName` according to the documentation in multiple places.

I'm out of ideas for today, but may be able to try again next week.

himerus:
I can confirm and am investigating at the moment. This is odd behavior. Good catch by setting up/updating the `dev` folder setup. I can validate this on a client project as I attempt to setup things under a client specific folder.

himerus:
Here is a working `rollup.config.js` file I'm using:

```javascript
import summary from 'rollup-plugin-summary';
import { terser } from 'rollup-plugin-terser';
import resolve from '@rollup/plugin-node-resolve';
import replace from '@rollup/plugin-replace';
import multi from '@rollup/plugin-multi-entry';
import typescript from 'rollup-plugin-typescript2';
import svg from 'rollup-plugin-svg';

export default {
  // Output all components to a single exported file.
  input: {
    include: [
      'src/components/**/*.ts',
    ],
    exclude: [
      'src/components/**/*.stories.ts',
      'src/components/**/*.css.lit.ts',
      'src/components/**/*.test.ts',
    ],
  },
  output: [
    {
      // Modern bundle.
      file: 'dist/outline.js',
      format: 'esm',
    },
    {
      // Legacy bundle.
      file: 'dist/outline-legacy.js',
      format: 'cjs',
    },
  ],
  onwarn(warning) {
    if (warning.code !== 'THIS_IS_UNDEFINED') {
      console.error(`(!) ${warning.message}`);
    }
  },
[...]

PR #80, opened 2021-07-27
Rollup configuration updates

damontgomery:
These should not be included.
They also make the package too large for a single file and cause errors / some things to not be included.

## Testing steps

```
yarn install
yarn rollup
```

Confirm you don't get errors with `yarn rollup`.

View the `dist/outline.js` file and search for `my-element`.

Could also test with #75.

## Notes

See #79 

This may not be a permanent fix, but it seems to work for now. If the repository becomes too large, I think we'll hit the same issue and need to see if there is another solution.

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/80"></a>

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -8,7 +8,16 @@ import svg from 'rollup-plugin-svg';
 
 export default {
   // Output all components to a single exported file.
-  input: 'src/components/**/*[!(stories|lit)].ts',
+  input: {
+    include: [
+      'src/components/**/*.ts',
+    ],
+    exclude: [
+      'src/components/**/*.stories.ts',
+      'src/components/**/*.css.lit.ts',
+      'src/components/**/*.test.ts',
+    ],
+  },
   output: [
     {
       // Modern bundle.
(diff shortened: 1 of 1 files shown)


himerus:
Eventually this will work with splitting the way our Stencil version worked. And it used Rollup under the hood, so it's just future tweaking to further refine Rollup. 

Thanks for getting this pushed up!

rollup.config.js
```diff
@@ -8,7 +8,14 @@ import svg from 'rollup-plugin-svg';
 
 export default {
   // Output all components to a single exported file.
-  input: 'src/components/**/*[!(stories|lit)].ts',
+  input: {
+    include: ['src/components/**/*.ts'],
+    exclude: [
+      'src/components/**/*.stories.ts',
+      'src/components/**/*.css.lit.ts',
+      'src/components/**/*.test.ts',
```
himerus, line 16: This worked for me. We will want to though in conjunction with other work most likely by default, exclude the code examples as we don't want to actually send out `extended-element`, etc. out to the consumer. They would still be visible in Storybook, but not part of the production build that Rollup powers. 

We can discuss that on #75 and what our examples on that page look like.

himerus (approved):

PR #94, opened 2021-08-06
list component.

LeeMellon:
### Adds a basic list component.
 - Handles vert/horiz alignment, and switches between mobile/desktop.
 - generates ul/ol/div lists
 - optional nav prop for generating navigation menus. 
 - optional header and footer slots
 - optional dividers in P2 colors.
 - optional column implementation from 2-4 cols. 
 
Cursor_and_Atoms___List_-_List_⋅_Storybook

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/94"></a>


mabry1985 (approved):
This looks good to me!

src/components/base/outline-list/outline-list.css
```diff
@@ -0,0 +1,118 @@
...
+ul.grid > ::slotted(li),
+ol.grid > ::slotted(li) {
+  @apply p-2 flex items-center whitespace-nowrap;
+}
+
+.columns--2,
+.columns--3,
+.columns--4 {
+  display: flex;
+  flex-direction: column;
```
himerus, line 24: I want to update some of this CSS to use `@apply` where appropriate/possible. This one would simply be `@apply flex flex-col;`. 

I would do this throughout this CSS file, including using some of the `demo` color palette for the colors.

himerus (dismissed):
Minor tweaks to use more Tailwind classes where possible. I may do this myself as cleanup before merging if you don't get to it in the next day or two.

PR #96, opened 2021-08-10
Feature/breadcrumbs

### LINKED ISSUE #42: Breadcrumb Component
## Requirements

Create a basic breadcrumb component that:

- Takes either a data array to render as links, or passes pre-rendered links into a slot.
- Conforms to accessibility guidelines.

### Example
https://designsystem.digital.gov/components/breadcrumb/
Breadcrumb___U_S__Web_Design_System__USWDS_

### Accessibility Guidelines 
[A11y Style Guide](https://a11y-style-guide.com/style-guide/section-navigation.html#kssref-navigation-breadcrumbs)
[W3.org](https://www.w3.org/TR/wai-aria-practices-1.1/examples/breadcrumb/index.html)

sarahp:
Improves #42.

## Testing Notes
1. Open Storybook
2. Go to Navigation at the bottom of the left side
3. Choose Breadcrumbs
4. Observe that there are links
5. Hover over the first one
6. Observe that you see the color change to green with an underline
7. Observe that the last link has no `/` always has an underline
8. Observe that as you hover down the line the `/` remain blue.

## Example

Screen Shot 2021-08-13 at 3 38 23 PM

## Notes

The markup is from [Drupal.org](https://www.drupal.org/association). I added the last and active classes to make it possible to change the behavior of those last two links. 

## Possible additional features

It would be great to have some variants of different types of dividers.

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/96"></a>


src/components/base/outline-breadcrumbs/outline-breadcrumbs.ts
```diff
@@ -0,0 +1,30 @@
+// Our base component, which all others extend.
+import { OutlineElement } from '../outline-element/outline-element';
+import { CSSResultGroup, html, TemplateResult } from 'lit';
+import { customElement } from 'lit/decorators.js';
+import componentStyles from './outline-breadcrumbs.css.lit';
+import { SlottedController } from '../../controllers/slotted-controller';
```
shaal, line 6: @sarahp `slotted-controller` is missing

LeeMellon: @shaal does the component need to be wrapped in a `<slotted-controller>`?

sarahp: @LeeMellon Without the slotted-controller, I couldn't style the individual links of the Drupal output.

LeeMellon: @sarahp I get that. I was wondering what @shaal was saying was missing.

LeeMellon: Oh NM now I see. You're importing it but it isn't there.

sarahp: It's in a gist because I thought Jake might be using it's own PR, but I could push it up if that's not the case: https://gist.github.com/sarahp/f23c222e8a01c214b7a95d5a249c41d1

src/components/base/outline-breadcrumbs/outline-breadcrumbs.css
```diff
@@ -0,0 +1,16 @@
+:host,
+a,
+::slotted(a) {
+  @apply font-body text-brand-primary transition-colors duration-300 mr-4 no-underline;
```
madhaze: @sarahp text-brand-primary seems like it shouldn't be applied to `:host`. So that the `/` isn't the link color.

src/components/base/outline-breadcrumbs/outline-breadcrumbs.ts
```diff
@@ -0,0 +1,35 @@
...
+export class OutlineBreadcrumbs extends OutlineElement {
+  static styles: CSSResultGroup = [
+    componentStyles,
+    css`
+      span.last::after {
+        display: none;
+      }
+      span::after {
+        content: '/';
+        margin: 0 15px 10px;
```
madhaze, line 24: the Left margin here should be removed or the `mr-4` on `host::, a, ::slotted(a)` since it looks to double up the margin.

madhaze (changes requested):

src/components/base/outline-breadcrumbs/outline-breadcrumbs.css
himerus: Removed references to `:host` and everything is still styled as expected.

src/components/base/outline-breadcrumbs/outline-breadcrumbs.ts
himerus: Removed margin and things look better: 

Screenshot of Google Chrome (9-9-21, 8-02-14 PM)

PR #101, opened 2021-08-16
Launch Gitpod interface even faster

shaal:
# :racing_car:
Now that Node version 16 is no longer required, we can Gitpod's default Node version 14 (active LTS).
This simplifies the process since a custom docker image is no longer required.

### :racing_car: _TLDR: Gitpod's interface loads faster._ :racing_car:

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/101"></a>


damontgomery:
It loads and storybook works.

Is 16 recommended even if not required? Is there value in running one version consistently on all environments (local / ci / Gitpod)?

damontgomery:
Since 14 is the first link on https://nodejs.org/en/ and is the default on many of these systems, I think it's OK to use that.

I think there is a benefit to having the local / ci / Gitpod environments all using the same version if possible.

https://github.com/phase2/outline/blob/next/.github/workflows/build.yml#L43

https://github.com/phase2/outline/blob/next/.nvmrc

himerus (approved):

grshane (approved):

PR #105, opened 2021-08-27
fix: imported horizontal alignment from outline elements into the con…

### LINKED ISSUE #90: Conflicting namespaces: ALL_HORIZONTAL_ALIGNMENT
```
(!) Conflicting namespaces: "virtual:multi-entry.js" re-exports "ALL_HORIZONTAL_ALIGNMENT" from both "src/components/base/outline-container/config.ts" and "src/components/base/outline-element/utils/types.ts" (will be ignored)
```

The instance in `src/components/base/outline-container/config.ts` should be removed and instead reference the shared one in `outline-element`.

tomsshow:
…tainer config to avoid namespace issue

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/105"></a>

### DIFF
--- a/src/components/base/outline-container/config.ts
+++ b/src/components/base/outline-container/config.ts
@@ -1,3 +1,4 @@
-export const ALL_HORIZONTAL_ALIGNMENT = ['left', 'center', 'right'] as const;
-type HorizontalAlignmentTuple = typeof ALL_HORIZONTAL_ALIGNMENT;
+import { ALL_HORIZONTAL_ALIGNMENT } from "../outline-element/utils/types";
+let HORIZONTAL_ALIGNMENT = ALL_HORIZONTAL_ALIGNMENT;
+type HorizontalAlignmentTuple = typeof HORIZONTAL_ALIGNMENT;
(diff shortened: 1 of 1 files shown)


src/components/base/outline-container/config.ts
```diff
@@ -1,3 +1,4 @@
-export const ALL_HORIZONTAL_ALIGNMENT = ['left', 'center', 'right'] as const;
-type HorizontalAlignmentTuple = typeof ALL_HORIZONTAL_ALIGNMENT;
+import { ALL_HORIZONTAL_ALIGNMENT } from "../outline-element/utils/types";
+let HORIZONTAL_ALIGNMENT = ALL_HORIZONTAL_ALIGNMENT;
```
madhaze, line 2: @tomsshow can you update this to `const` as per the linting issue.

madhaze (changes requested):

himerus (changes requested):
Please follow the instructions in #90. 

@tomsshow This approach needs to be updated.

himerus:
Closing in favor of #133 That approach fixes the referenced issue appropriately.

PR #106, opened 2021-09-10
Adding core Shoelace utilities

himerus:
Many of these will be required by various components we pull in from Shoelace. 
Also, the utilities are generally solid, and useful for other uses we may come up with. 

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/106"></a>

### DIFF
--- a/src/utilities/animate.ts
+++ b/src/utilities/animate.ts
@@ -1,0 +1,85 @@
+//
+// Animates an element using keyframes. Returns a promise that resolves after the animation completes or gets canceled.
+//
+export function animateTo(
+  el: HTMLElement,
+  keyframes: Keyframe[],
+  options?: KeyframeAnimationOptions
+) {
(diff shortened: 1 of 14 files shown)


damontgomery:
Can you add comments on the files or to a readme that explains the origin? May help if there are issues and someone wants to see if there are upstream changes.

himerus:
> Can you add comments on the files or to a readme that explains the origin? May help if there are issues and someone wants to see if there are upstream changes.

@damontgomery Are you referring to just the original origin (Shoelace)? I felt like the utility functions were fairly self-documenting. 

If you were thinking a README that outlined some basics of what was there and what some of it did (or at least what the functions were), I could get on board with that.

himerus:
Gah. Nevermind. I read your comment more closely. 

> May help if there are issues and someone wants to see if there are upstream changes.

I agree.

PR #118, opened 2021-10-06
Feature/outline icon update

LeeMellon:
Update to match @mabry1985 work on internal project

Cursor_and_Branding___Graphics___Icons_-_Icons_⋅_Storybook

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/118"></a>


src/components/base/outline-accordion/outline-accordion.css
```diff
@@ -1,19 +1,16 @@
...
 }
 
-.accordion-panel {
-  border: black solid 2px;
-  border-top: none;
-}
-.accordion-heading {
-  @apply py-2 px-4;
+.accordion-title.mobile {
+  @apply text-blue-700 text-lg;
```
himerus: Let's use the `demo` palette here as well. So maybe `text-demo-darkBlue`. 
We need to expand the `demo` palette to have some more colors for that out of the box styling.

himerus (changes requested):

src/components/base/outline-accordion/outline-accordion.css
himerus: Pushed fix.

damontgomery:
Cool. :) Where did the icons come from / are they free to use on any project / have rules around usage?

We are using Google Material Icons on another internal project and have adapted and earlier version of this work to handle those icons.

PR #124, opened 2021-10-20
fix(concurrently): adding kill-others to concurrently scripts

himerus:
Solves "once and for all" an issue where when using `concurrently` in our `yarn start` script, the `generate.css` script was failing out during an early issue, but it appeared the watch script was still running appropriately unless you saw the exit code for `generate.css` in the logs.


damontgomery:
Current behavior is that the script exits when a class is not found.

You can test this by adding the following to any component `.css` file.

```
* {
  @apply not-a-tailwind-class;
}
```

PR #129, opened 2021-10-20
Youtube video component

damontgomery:
This provides a way to embed a youtube video that assumes 16/9 ratio and scales to fit the space provided.

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/129"></a>


himerus:
This is awesome. And incredibly simple. Worth inclusion for sure. I updated the default video to be a P2 one. 

Somehow we need to put out some content so we can point to an Outline video!

PR #136, opened 2021-10-25
Vimeo Video component

damontgomery:
Basic Vimeo video component that fits the available space.

Uses the same structure as Youtube Video.

If we want to use `aspect-ratio` the CSS can be

```css
iframe {
  @apply block w-full;

  aspect-ratio: 16/9;
}
```

We used this on an internal project and seems to work OK.

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/136"></a>

### DIFF
--- a/src/components/base/outline-video-vimeo/outline-video-vimeo.stories.ts
+++ b/src/components/base/outline-video-vimeo/outline-video-vimeo.stories.ts
@@ -1,0 +1,52 @@
+import { html, TemplateResult } from 'lit';
+import './outline-video-vimeo';
+import { ifDefined } from 'lit/directives/if-defined.js';
+
+const configuration = {
+  title: 'Content/Vimeo Video',
+  component: 'outline-video-vimeo',
+  argTypes: {
+    videoID: {
+      name: 'video-id',
+      description: 'Vimeo video id',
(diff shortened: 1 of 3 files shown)


src/components/base/outline-video-vimeo/outline-video-vimeo.stories.ts
```diff
@@ -0,0 +1,52 @@
...
+    },
+  },
+};
+
+export default configuration;
+
+const Template = (args = configuration.args): TemplateResult => {
+  args = {
+    ...configuration.args,
+    ...args,
```
danmuzyka, line 43: What other remaining args might we have here beyond what was in `configuration.args`? It doesn't look as though the component is built to accept any other args.

damontgomery: @danmuzyka. Good question. :)

This general pattern is described in https://github.com/phase2/outline/discussions/109. It's what I was able to settle on to use stories in stories to avoid copy & paste and refactor challenges.

Happy to discuss more either on a call or in that thread if it's helpful.

himerus (approved):

danmuzyka (approved):

PR #137, opened 2021-10-27
Feature/outline image slider

### LINKED ISSUE #107: Slider Component
We need a slider. 
An accessible slider. 
Maybe more than one slider?
Inspiration links pending.

mabry1985:
- Adds Splidejs Carousel library
- Adds `outline-image-slider`
- Closes [Issue 107](https://github.com/phase2/outline/issues/107)

Mobile: 
Screen Shot 2021-10-27 at 4 49 35 PM

Full Viewport: 
Screen Shot 2021-10-27 at 4 51 14 PM

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/137"></a>


src/components/base/outline-image-gallery/outline-image-gallery.css
```diff
@@ -0,0 +1,112 @@
+@import '@splidejs/splide/dist/css/splide-core.min.css';
```
mabry1985: @shaal Here is how I ended up importing the css from node modules

shaal: @mabry1985 nice!!!

src/components/base/outline-image-slider/outline-image-slider.css
```diff
@@ -0,0 +1,111 @@
...
+.splide {
+  @apply relative;
+}
+
+.splide__pagination {
+  display: none;
+}
+
+.splide__arrows {
+  @apply bg-brand-primary absolute flex justify-center bottom-0 left-0 z-40 w-full;
```
grshane: For extensibility would it make sense to define a css var for the arrow color?

grshane (approved):

src/components/base/outline-image-slider/outline-image-slider.css
mabry1985: Sorry I missed this comment when it was first added. Yeah I think that makes sense, there would be more follow up tickets needed to make it a little more flexible for sure.

src/components/base/outline-image-slider/outline-image-slider.css
```diff
@@ -0,0 +1,111 @@
+@import '@splidejs/splide/dist/css/splide-core.min.css';
+
+.splide {
+  @apply relative;
+}
+
+.splide__pagination {
+  display: none;
+}
```
damontgomery, line 9: Splide has a configuration to disable pagination. We're using this on an internal project if you want to use that. Not necessary.

src/components/base/outline-image-slider/outline-image-slider.css
```diff
@@ -0,0 +1,111 @@
...
+  border-radius: 0;
+  height: 12rem !important;
+  width: 100% !important;
+}
+
+.splide__slide.is-visible {
+  visibility: visible;
+}
+
+.sr-only {
```
damontgomery, line 62: minor: Can we call this `screen-reader-only`?

damontgomery (commented):
Thanks, @mabry1985!

I used this as a model for an internal project and it worked well.

I can understand the code.

For general purpose use, I'm not sure what to make of this type of component. I had to modify the Splide configuration and CSS a lot for our use on that project since the carousel is quite different. So, it's a nice starting point, but I can't imagine a project using it as-is.

I also took a different approach where I copied the slides into the shadow DOM instead of using dynamically created slots per slide. That to me is a little simpler, but this is clear enough and works well.

Thanks again and happy to talk through that other example if it's helpful.

src/components/base/outline-image-slider/outline-image-slider.css
mabry1985: Yeah I'm aware of the Splide configuration there, I believe this was how I saw it done in the accessible boilerplate to create your own buttons while still using the functionality of the pagination. I think it's probably redundant though. Good catch.

src/components/base/outline-image-slider/outline-image-slider.css
mabry1985: sr-only is what I've seen it called in several places, but it's no matter to me what we call it here.

mabry1985:
@damontgomery thanks for taking a look and the thoughtful pr. I'd be happy to take a look at your iteration on this. I was just looking to back port the component from SMC and we could add any follow up tickets needed to improve it. We discussed several weeks ago about getting stuff in quick and dirty and improving when there is time. I don't know if this means leaving these components in the PR queue until there is time to update them or pulling them into a branch of their own. @himerus what are your thoughts there?

himerus (changes requested):
@mabry1985 Can you look at https://github.com/phase2/outline/security/dependabot/yarn.lock/prismjs/open and see if we can bump the version of splide enough to avoid this error?

NEVERMIND. This was related to something else anyway!

himerus (approved):

himerus:
Awaiting fix for caption and navigation positioning.

himerus:
I plan to pick this one up likely quite soon. We used this component as a starting point for another project, and I have some updates to pull in as well as the adjustments that may need to be pulled back in from the original work.

PR #140, opened 2021-11-03
Updates to genearate-icons.js

sarahp:
We updated the generate-icons.js, this is before Jake and Tom ran into some issues with it. Here's what we got to work for generating new icons for Outline.

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/140"></a>

Edit: We added a few icons to support future porting of some of the Infographic components from Michigan Medicine.


scripts/generate-icons.js
```diff
@@ -73,7 +75,7 @@ const generateSvgImports = svg => {
 
 const generateIconList = svg => {
   return Object.keys(svg).map(key => {
-    if (key.includes('-')) {
+    if (nameDelimiter.some(i => key.includes(i))) {
```
damontgomery, line 78: Would we be able to reduce duplication of the list of character replacements if we ran `camelize` all the time without checking for a list of characters?

Or, could the logic be grouped together, for example, putting the list of characters and conditional checks inside `camelize`?

Right now, updates require changes in the `camelize` function as well as the `nameDelimiter` array. These two are defined quite far apart in the code and then used in a third place: lines 15, 51, and 78.

damontgomery (commented):
Thanks.

I didn't review the functionality itself, but the code changes seem mostly clear.

Could we make a change to just `generate-icons.js` here? It looks like `SVGO` configuration was also changed and it's not clear to me if it should / should not change. Similarly, there are some changes to the list of icons included.

The description says there were some issues with projects. Were those when they added their own icons? Can we add more testing steps here to try and reproduce the issue / fix?

Thanks.

scripts/generate-icons.js
mabry1985: Yeah grouping the logic makes sense to me

himerus (approved):

danmuzyka (approved):

PR #141, opened 2021-11-07
Fix(slotted controller) edge cases

shaal:
# The Problem/Issue/Bug
When a web component that utilize slotted-controller is placed inside another web component, slotted-controller functionality breaks, and the slotted content is not displayed.

## How this PR Solves the Problem
This PR resolves additional edge cases for moving slotted content from Light DOM to Shadow DOM:
- Improved the way it finds accurately slotted content.
- Allow processing both named-slots and default-slots
- Doesn't break when there is no slotted content.
- and more :)

# Additional details

## Manual Testing Instructions
* `yarn start`
* Open this URL in browser: http://localhost:6006/?path=/docs/molecules-card--card-with-breadcrumb
* Inspect the breadcrumbs component inside the card component
* Confirm <slot></slot> is left empty
* Confirm all content is displayed in the Shadow DOM in its original order as it would in Light DOM.

## Release/Deployment notes

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/141"></a>

### DIFF
--- a/src/components/controllers/slotted-controller.ts
+++ b/src/components/controllers/slotted-controller.ts
@@ -19,19 +19,40 @@ export class SlottedController implements ReactiveController {
   }
 
   onSlotChange() {
-    const elementName = this.hostEl.tagName;
-    const myElement: HTMLElement | null = this.hostEl.parentElement
-      ? this.hostEl.parentElement.querySelector(elementName)
-      : null;
-    if (myElement && myElement.shadowRoot) {
-      const slot: HTMLElement | null = myElement.shadowRoot.querySelector(
-        'slot'
-      )
-        ? myElement.shadowRoot.querySelector('slot')
-        : null;
-      if (slot) {
-        slot.append(...myElement.children);
-      }
+    // Search for all slots provided by the Web Component.
+    const slotsArray = this.hostEl.renderRoot.querySelectorAll('slot');
+
+    // If any slots found
+    if (slotsArray.length > 0) {
+      const currentElement = this.hostEl;
+      // Go through each slot shadow dom position
+      slotsArray.forEach(slot => {
+        // Process named slot
+        if (slot.name) {
+          // Fetch the corresponding named slot in the lightDom
+          const slotLightDom = currentElement.querySelector(
+            '[slot=' + slot.name + ']'
+          );
+          // Only if named slot found in light dom - move it into shadow DOM
+          if (slotLightDom) {
+            slot.before(slotLightDom);
+          }
+        }
+        // Process default (unnamed) slot
+        else {
+          // Get all content that doesn't have slot as attribute
+          const slotLightDomArray = Array.from(currentElement.children).filter(
+            node => {
+              return !node.getAttributeNode('slot');
+            }
+          );
+
+          // Move all unnamed slot content to the corresponding position in shadow DOM
+          slotLightDomArray.forEach(slotLightDom => {
+            slot.before(slotLightDom);
+          });
+        }
+      });
     }
   }
 }
--- a/src/components/controllers/slotted-controller.ts
+++ b/src/components/controllers/slotted-controller.ts
@@ -1,12 +1,16 @@
-import { ReactiveControllerHost, ReactiveController } from 'lit';
+import {
+  ReactiveControllerHost,
+  ReactiveController,
+  ReactiveElement,
+} from 'lit';
 
 export class SlottedController implements ReactiveController {
   host: ReactiveControllerHost;
-  hostEl: HTMLElement;
+  hostEl: ReactiveElement;
 
   constructor(host: ReactiveControllerHost) {
     (this.host = host).addController(this);
-    this.hostEl = this.host as unknown as HTMLElement;
+    this.hostEl = this.host as unknown as ReactiveElement;
   }
 
   hostConnected() {
--- a/src/components/base/outline-breadcrumbs/outline-breadcrumbs.css
+++ b/src/components/base/outline-breadcrumbs/outline-breadcrumbs.css
@@ -1,14 +1,11 @@
-a,
-::slotted(a) {
+a {
   @apply font-body text-demo-darkBlue transition-colors duration-300 no-underline;
 }
 
-a:hover,
-::slotted(a:hover) {
+a:hover {
   @apply text-demo-mediumBlue underline;
 }
 
-a.active,
-::slotted(a.active) {
+a.active {
   @apply underline;
 }
--- a/src/components/base/outline-card/outline-card.stories.ts
+++ b/src/components/base/outline-card/outline-card.stories.ts
@@ -249,3 +249,31 @@ CardWithHeaderAndFooter.args = {
     </div>
   `,
 };
+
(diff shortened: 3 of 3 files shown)


himerus (approved):
Approved! Moving to a client project now to test thoroughly.

src/components/controllers/slotted-controller.ts
```diff
@@ -19,19 +23,40 @@ export class SlottedController implements ReactiveController {
...
+            }
+          );
+
+          // Move all unnamed slot content to the corresponding position in shadow DOM
+          slotLightDomArray.forEach(slotLightDom => {
+            slot.before(slotLightDom);
+          });
+        }
+      });
     }
```
chrissnyder2337, line 60: return early for easier to read code and less nesting.

```suggestion
    // Nothing to do if we have no slots.
    if (!slotsArray.length) {
      return;
    }
    
    const currentElement = this.hostEl;
    // Go through each slot shadow dom position
    slotsArray.forEach(slot => {
      // Process named slot
      if (slot.name) {
        // Fetch the corresponding named slot in the lightDom
        const slotLightDom = currentElement.querySelector(
          '[slot=' + slot.name + ']'
        );
        // Only if named slot found in light dom - move it into shadow DOM
        if (slotLightDom) {
          slot.before(slotLightDom);
        }
      }
      // Process default (unnamed) slot
      else {
        // Get all content that doesn't have slot as attribute
        const slotLightDomArray = Array.from(currentElement.children).filter(
          node => {
            return !node.getAttributeNode('slot');
          }
        );

        // Move all unnamed slot content to the corresponding position in shadow DOM
        slotLightDomArray.forEach(slotLightDom => {
          slot.before(slotLightDom);
        });
      }
    });
```

src/components/controllers/slotted-controller.ts
```diff
@@ -1,12 +1,16 @@
...
-  hostEl: HTMLElement;
+  hostEl: ReactiveElement;
 
   constructor(host: ReactiveControllerHost) {
     (this.host = host).addController(this);
-    this.hostEl = this.host as unknown as HTMLElement;
+    this.hostEl = this.host as unknown as ReactiveElement;
   }
 
   hostConnected() {
```
damontgomery, line 16: Can you listen to `slotchange` on a web component? I thought this event only happens on the slotted elements themselves. I could be wrong.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement/slotchange_event

If you can't, this code just doesn't ever fire.

src/components/controllers/slotted-controller.ts
```diff
@@ -19,19 +23,40 @@ export class SlottedController implements ReactiveController {
...
-      const slot: HTMLElement | null = myElement.shadowRoot.querySelector(
-        'slot'
-      )
-        ? myElement.shadowRoot.querySelector('slot')
-        : null;
-      if (slot) {
-        slot.append(...myElement.children);
-      }
+    // Search for all slots provided by the Web Component.
+    const slotsArray = this.hostEl.renderRoot.querySelectorAll('slot');
```
damontgomery, line 27: Is there a better name for this or can you convert this to an Array?

https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll

This is actually a `NodeList` of `Element`.

Why not just `slots` or `slotElements`?

src/components/controllers/slotted-controller.ts
```diff
@@ -19,19 +23,40 @@ export class SlottedController implements ReactiveController {
...
+
+    // If any slots found
+    if (slotsArray.length > 0) {
```
damontgomery, line 30: Why check the length and then run a `foreach`? The foreach will do nothing if you have no items, so I don't think you need to check the length first.

src/components/controllers/slotted-controller.ts
```diff
@@ -19,19 +23,40 @@ export class SlottedController implements ReactiveController {
...
+      const currentElement = this.hostEl;
```
damontgomery, line 31: Why rename the class property here? Can you call `this.currentElement` if you like that name better?

src/components/controllers/slotted-controller.ts
```diff
@@ -19,19 +23,40 @@ export class SlottedController implements ReactiveController {
...
+      // Go through each slot shadow dom position
+      slotsArray.forEach(slot => {
```
damontgomery, line 33: Could you make a set of functions?

`copyLightDOMElementsIntoShadowDOM(slot): null`
`getLightDOMElements(slotName): Elements[]`
`prependLightDOMElements(slotName, elements): null`


```javascript
slots.forEach(slot => {copyLightDOMElementsIntoShadowDOM(slot)});

function copyLightDOMElementsIntoShadowDOM(slot) {
  prependLightDOMElements(slot.name, getLightDOMElements(slot.name));
}

function getLightDOMElements(slotName) {
  if (isDefaultSlot) {
  
  }
  
  if (isDefaultSlot === false) {

  }
}

function prependLightDOMElements(elements) {
  ...
}
```

damontgomery (commented):
Overall, I can read the code and it makes sense!

I left some comments about the specific names / structure based on our discussion today, if that is interesting.

I didn't test the code myself.

Thanks.

src/components/controllers/slotted-controller.ts
mabry1985: @damontgomery You are correct, you have to attach the event to the slotted element itself. I tried to do this on a component for SMC and I can confirm it does not work.

shaal: @mabry1985 can you add more details of what works or what doesn't work?
(I don't think I added this part of the code, so I am not familiar with it)

damontgomery: `slotchange` events fire on `slot` elements and don't bubble. So, you can add `@slotchange=...` But the web component (parent) doesn't get those events.

So, if you want to respond to slot changes, you need to watch all the slots. Or, you can add a mutation observer on the light DOM element.

```javascript
  observer: MutationObserver;

  connectedCallback() {
    super.connectedCallback();

    this.observer = new MutationObserver(() => {
      this.requestUpdate();
    });
    this.observer.observe(this, { childList: true });
  }

  disconnectedCallback() {
    this.observer.disconnect();
  }
```

I haven't done this on a Lit controller, so I'm not sure if that works, but it does on a component.

src/components/controllers/slotted-controller.ts
himerus: Agreed on clarification of variable name.

src/components/controllers/slotted-controller.ts
himerus: Anyone have time to dig into this a little more? This sounds interesting.

himerus:
Discussion on **12022021**: Hold this PR until a few items are addressed.

damontgomery:
Specific blocker to merge:
- Remove the slot change handler since it doesn't do anything currently

Nice to have:
- Get slot change handlers to work if possible so that any slot changes will cause a component update

himerus:
Closing in favor of https://github.com/phase2/outline/pull/203

PR #191, opened 2021-12-13
Updates for outline-container & outline-grid

mabry1985:
## Description

- Fix issue with `is-nested` where the value could never be set to false.
- Refactor `is-nested` to be `x-padding` to be more descriptive.
- Feature: add `y-padding`
- Configure these values via CSS Custom Properties. 
- Update the `outline-grid` to properly utilize the new `x-padding` and `y-padding` properties from `outline-container`
- Fix multiple issues with `outline-grid` Storybook stories.
- Fix issues where storybook was unable to properly display the `xxxl` viewport (2160px) causing "oddities" in the display at previous max-width. This was needed to properly test the container/grid at ALL breakpoints.
- Adds additional descriptions and default value labels in container/grid stories.

Creates CSS Variables to handle the spacing for the x/y padding on default containers. Typically we've NOT used the `y-padding`, allowing the content inside a container to handle this, however, I can see reason this could be useful. 

```css
 /* 
  * Configuration values for `outline-container`.
  */
  --outline-container-padding-x: 1.5rem; 
  --outline-container-padding-y: 1.5rem;
```

## Testing

- View `Container`  and `Grid` in storybook and toggle the new padding controls.

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/191"></a>

### DIFF
--- a/src/components/base/outline-container/outline-container.css
+++ b/src/components/base/outline-container/outline-container.css
@@ -1,42 +1,8 @@
 :host {
-  @apply block px-6 max-w-full-screen-w font-body;
+  @apply block p-6 max-w-full-screen-w font-body;
+  @apply sm:max-w-screen-sm md:max-w-screen-md lg:max-w-screen-lg xl:max-w-screen-xl xxl:max-w-screen-xxl xxxl:max-w-screen-xxxl;
 }
 
-@screen sm {
-  :host {
-    @apply max-w-screen-sm;
-  }
-}
-
-@screen md {
-  :host {
-    @apply max-w-screen-md;
-  }
-}
-
-@screen lg {
-  :host {
-    @apply max-w-screen-lg;
-  }
-}
-
-@screen xl {
-  :host {
-    @apply max-w-screen-xl;
-  }
-}
-
-@screen xxl {
-  :host {
-    @apply max-w-screen-xxl;
-  }
-}
-
-@screen xxl {
-  :host {
-    @apply max-w-screen-xxxl;
-  }
-}
 :host(:not([is-nested])) {
   @apply px-0;
 }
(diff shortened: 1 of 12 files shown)


damontgomery:
Thanks for looking into this, @mabry1985.

I made a PR against this branch in #192 while reviewing. I added some comments about why I went down that path.

With this PR, I wasn't able to see the changes reflected in Storybook when I changed the controls.

damontgomery:
Not sure why the coding standards is failing all of a sudden. I can try to take a look.

damontgomery:
Updated title / description since it didn't match any longer.

src/components/base/outline-container/outline-container.css
```diff
@@ -37,62 +37,54 @@
     @apply max-w-screen-xxxl;
   }
 }
-:host([is-nested]),
-:host([is-nested='true']) {
+:host(:not([is-nested])) {
```
damontgomery: This naming is confusing, but the current implementation matches the Storybook name `Padding` so that when you set `Padding` to `true`, there is padding.

If that isn't the intent, we may want to change the label in Storybook alongside this CSS rule.

himerus: This is now updated to `x-padding`.

himerus (dismissed):

damontgomery (approved):
Tested `Container` and `Grid` stories in Gitpod and controls make sense and work as expected.

- Testing notes were very helpful.
- Naming is consistent with component and make sense.
- CSS for `outline-container` seems more straight forward and makes sense.
- Didn't run automated tests since they don't seem to work in Gitpod.
  - `browserType.launch: Executable doesn't exist at /home/gitpod/.cache/ms-playwright/chromium-920619/chrome-linux/chrome`.

Wasn't aware of adding columns to the stories like with the default values. Cool to see.

Thanks!

grshane (approved):
I agree the testing step were helpful

PR #198, opened 2021-12-22
[DRAFT] fix: adds return to setactive for if there is no id on the target

mabry1985:
* Currently if the accordion is set to `single-panel = true` then you are able to click on the panel itself to close it, which is not desired behavior. For instance if there is something in that panel that you are wanting to interact with then it will close the panel.

* This behavior is happening because the setActive click handler is running the close logic even when the target element has no id, such as in the case when you click on the panel itself. 
* This PR adds a return if there is not id on the element, preventing the unwanted behavior. This still works as desired when clicking on the heading to close the panel.

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/198"></a>

### DIFF
--- a/src/components/base/outline-accordion/outline-accordion.ts
+++ b/src/components/base/outline-accordion/outline-accordion.ts
@@ -82,6 +82,8 @@ export class OutlineAccordion extends OutlineElement {
     const element = e?.target as HTMLElement;
     const contentId = element.id;
 
+    if (!contentId) return;
+
     // if single-panel = true
 
     if (this.singlePanel) {


damontgomery (commented):
Thanks, @mabry1985!

I tested on Gitpod and I think this works mostly as you described, but I'm not sure.

- What can I click to close the accordion?
  - Right now I can click on the side or the top bar, is this intended?
- Would a code comment be helpful?
- Is this check sufficient?

## What can I click to close the accordion?

You can also click on the side and trigger the close:

click_area

## Would a code comment be helpful?

I think your comments in the PR are helpful and would help clarify what this code is doing.

It was not immediately obvious to me why we need to see if the ID exists on the target. So, a comment about how we want to distinguish clicks on bars rather than the whole panel might be helpful.

The comment I think would also help if this breaks in the future.

`isTargetAccordionButton = ...`?

## Is this test sufficient?

Right now, I'm not sure that this check is sufficient since changes to the panel or different types of slotted content could fail the check.

Here are some IDs and elements that I saw during testing:

elements_and_IDs

While the button / panel has an ID, anything slotted there may also have an ID.

Conceptually, I'm not sure of a simple fix.

The public interface of the panel doesn't have any methods such as emitting a `click` event when the target (button?) is clicked. So we would need to know about the internals to determine classes / IDs to use. If we had the panel emit the event, the consumer (accordion) doesn't need to know the internal structure of the panel, whether it has IDs, what those are, etc. It just needs to add a handler for the `click` event that the panel emits.

If this quick fix works as expected, maybe we can add a comment for future developers and look at the event as a nice-to-have?

## References
See https://github.com/phase2/outline/blob/next/src/components/base/outline-accordion-panel/outline-accordion-panel.ts.

mabry1985:
@damontgomery thank you for testing this, you found some stuff and brought up some points I hadn't considered. I will be setting this to draft and reassessing this PR.

mabry1985:
Outdated

PR #203, opened 2022-01-06
New & Improved Slot Controller

himerus:
## Description
This PR implements a [Reactive Controller](https://lit.dev/docs/composition/controllers/) intended to allow for better control over slots in a variety of scenarios. 

This code is based off of original work in the following locations:
- The original Outline [SlottedController](https://github.com/phase2/outline/blob/next/src/components/controllers/slotted-controller.ts)
- [Shoelace](https://github.com/shoelace-style/shoelace/blob/next/src/internal/slot.ts)

Supersedes #141 

## Implementations
Currently, there are two methods we'd use when implementing the `SlotController`. 

### General slot helpers, leaving Light DOM intact
This provides all the methods described below, and Outline will not do any manipulation of Light DOM nodes. 
```typescript
slots = new SlotController(
  this, // This, the host element.
  false // Do not manipulate Light DOM nodes.
);
```

### General slot helpers, transfer Light DOM to Shadow DOM
This provides all the methods described below, and Outline will move all Light DOM nodes to the Shadow DOM.
```typescript
slots = new SlotController(
  this, // This, the host element.
  true // Move all Light DOM nodes to the Shadow DOM.
);
```

## Methods
The following methods are defined with this controller. 

#### `hasDefaultSlot()`
Private method to determine if nodes exist inside the unnamed slot. 

#### `hasNamedSlot(name: string)`
Private method to determine if a named slot called `name` exists. 

#### `test(slotName?: string | undefined)`
Public method that can be utilized to check for the existence of all slots.

#### `getSlots()`
Public method to return an array of named slot names.

#### `moveNamedSlot(slotName: string)`
Private method that will move a specific named slot into the Shadow DOM.

#### `moveDefaultSlot()`
Private method to move all the nodes inside the default slot into the Shadow DOM.

#### `moveSlots()`
Private method that iterates and handles moving both named and unnamed slot content. This calls both `moveDefaultSlot()` and `moveNamedSlot()` for each named slot. Only enabled if `shadowShift` is set as `true`. 

## Testing
- Open GitPod
- Check the Breadcrumb component
- See that it looks the same as it did before at https://outline.phase2tech.com/?path=/docs/navigation-breadcrumbs--page
- Currently also used as a replacement for the original `SlottedController` in about 12 components on another internal project. The link to that PR can be seen in the history if you have appropriate Phase2 permissions. 

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/203"></a>


damontgomery (commented):
This looks cool, @himerus.

Could you add a comment at the top of the file with a link to the original in case that gets updates and we want to pull in the changes?

The controller itself has easy-enough to follow method names, but a brief comment at the top would be nice too to explain the general purpose / when to use.

src/components/controllers/slot-controller.ts
```diff
@@ -53,22 +53,22 @@ export class HasSlotController implements ReactiveController {
...
+/**
+ * Given a slot, this function iterates over all of its assigned element and text nodes and returns the concatenated
+ * HTML as a string. This is useful because we can't use slot.innerHTML as an alternative.
+ */
 export function getInnerHTML(slot: HTMLSlotElement): string {
   const nodes = slot.assignedNodes({ flatten: true });
   let html = '';
 
-  [...nodes].map(node => {
+  [...nodes].forEach(node => {
```
sarahp: Out of curiosity why did you switch from Map to ForEach?

sarahp (approved):
Overall this looks great! My only question has more to do with learning than anything else. Great work!

src/components/controllers/slot-controller.ts
himerus: In this instance, I actually can't tell you why. This was updated directly from Shoelace, and a couple of its utility functions. 

My guess would be it proved to be more performant in this instance.. But that's literally a guess.

src/components/controllers/slot-controller.ts
```diff
@@ -0,0 +1,189 @@
...
+  handleSlotChange(event: Event) {
+    const slot = event.target as HTMLSlotElement;
+
+    if (
+      (this.defaultSlot && !slot.name) ||
+      (slot.name && this.slotNames.includes(slot.name))
+    ) {
+      this.host.requestUpdate();
+    }
+  }
```
himerus: I can verify that adding new content to the slot doesn't bring that new element into the ShadowDOM. 

image

This `handleSlotChange` method will need some additional tweaking and testing.

image

himerus: This is now fixed. Just needed to add `moveSlots()` to the slot change handler.

PR #207, opened 2022-01-12
Update rules for commitlint/conventional

himerus:
The following configuration in `commitlint.config.js` on any existing projects will disable the checks for starting the subject with an uppercase letter, as well as allow for a full stop at the end. 

```javascript
module.exports = {
  extends: ['@commitlint/config-conventional'],
  rules: {
    // Make sure there is never a max-line-length by disabling the rule
    'body-max-line-length': [0, 'always'],
    // @see https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#subject-case
    'subject-case': [1, 'always', 'sentence-case'],
    // @see https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional#subject-full-stop
    'subject-full-stop': [1, 'always', '.'],
  },
};

```

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/207"></a>


damontgomery (approved):
Documentation is clear. I think this would be a good improvement.

I like to write in sentence case.

Although I hear the young kids don't do that any more.

grshane (approved):

PR #215, opened 2022-02-04
fix: add the ability to configure excluded stories

grshane:


src/.storybook/main.js
```diff
@@ -1,4 +1,13 @@
 const path = require('path');
+const outlineConfig = require('../../outline.config');
+
+const excludedStories = outlineConfig.excludedStories;
+function getExcluded() {
+  if ((excludedStories.length = 1)) {
```
damontgomery: What does this do? When would this fail?

outline.config.js
```diff
@@ -29,4 +29,7 @@ module.exports = {
     // default video in any storybook component.
     defaultVideo: 'xiqgG8HUZXE',
   },
+  // Exclude stories by adding the directory they can be found in
+  // ex ['outline-button'] this excludes any story in this directory
```
damontgomery, line 33: Is this excluding the directories or the files that match? My globbing may be getting confused. :) It looks like it looks for any story files that match the string (outline-button)*.stories.@(js|ts|mdx) in any folder rather than any story files in those folders.

Would that be `../components/**/!(${getExcluded()})/*.stories.@(js|ts|mdx)` with the slash before `*.stories`?

Glob_Tool___DigitalOcean

Glob_Tool___DigitalOcean


https://www.digitalocean.com/community/tools/glob

damontgomery (commented):
Thanks, @grshane.

Overall, this looks good and I can see it being helpful!

I left a few small comments. Some things were a little unclear. Could you take a look?

Thanks again.

src/.storybook/main.js
grshane: This prevents the | from being added if you're only excluding one story.

outline.config.js
grshane: Storybook uses micromatch not glob for their globbing structure, so that's why this syntax looks that way.

damontgomery (approved):
I haven't tested this on a project, but I think the options are clear and the code would be understandable if I needed to make changes.

The feature seems useful and I know it's been used on projects.

Thumbs up!

sarahp (approved):
This looks solid! Excited to have this added.

himerus (approved):

PR #217, opened 2022-02-09
chore: add prettier-tailwind plugin and format all css files

grshane:


shaal:
This PR is a great example of why using prettier with Tailwind makes no sense.

Take a look at all the files that changed. How does that help?

I would love to see prettier with regular CSS, when each css rule has its own line.
When you change margin for example, you would see the update just in margin.

sarahp (approved):
Looks good to me!

src/components/base/outline-card/outline-card.ts
```diff
@@ -80,6 +80,9 @@ export class OutlineCard extends OutlineElement {
   }
 
   render(): TemplateResult {
+    console.warn('warn');
+    console.error('warn');
+    console.log('warn');
```
himerus: What's up here?

himerus (changes requested):
I'm curious, where do we define that we are USING the plugin? Is there no change to the tailwind config? I apologize, I didn't read the article that talked about the new feature.

grshane:
> This plugin follows Prettier’s autoloading convention, so as long as you’ve got Prettier set up in your project, it’ll start working automatically as soon as it’s installed.

himerus (approved):

himerus:
image

As a note, if you use multiple `@apply` directives to separate concerns as shown above, this plugin will only format them PER line. 🔥 🔥 🔥

himerus:
> This PR is a great example of why using prettier with Tailwind makes no sense.
> 
> Take a look at all the files that changed. How does that help?

This helps by providing consistency with our current tools. That is all.

PR #227, opened 2022-02-21
fix: update Boolean to boolean

shaal:
TIL: `Boolean` vs `boolean`.
https://www.javascripttutorial.net/javascript-boolean/

This PR updates every occurrence of `Boolean` with `boolean`

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/227"></a>


damontgomery (approved):
Thanks for the reference. This makes sense to me.

himerus (approved):

PR #232, opened 2022-02-25
feat: add pre-push hook to lint and compile TS

grshane:


himerus:
Before we merge this, I want to discuss Husky's git hooks in general. 
Curious about some of the "holes" like just committing code when you haven't ran `yarn install`. Pipelines would catch things, but is there another way?

PR #233, opened 2022-02-25
Run axe-based a11y tests using storybook stories

shaal:
Utilizing this nice library https://github.com/chanzuckerberg/axe-storybook-testing

How to test -
Open this PR in Gitpod 💸 (there should be a button)
_You can stop the process in the terminal (which runs the default `yarn start`)_
Type in terminal: `yarn storybook.axe`

It will first build Storybook, and then run axe accessibility report on all the stories.

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/233"></a>


package.json
```diff
@@ -35,6 +35,7 @@
     "wds": "wds --app-index dev/index.html --watch --node-resolve",
     "preserve": "yarn prep",
     "serve": "concurrently -c \"cyan,blue\" \"yarn:rollup.watch\" \"yarn:wds\" --kill-others",
+    "storybook:axe": "yarn storybook.prod && yarn axe-storybook --build-dir=storybook-static/",
```
himerus: `storybook.axe` would follow the pattern below.

himerus (changes requested):

package.json
shaal:
```suggestion
    "storybook.axe": "yarn storybook.prod && yarn axe-storybook --build-dir=storybook-static/",
```

shaal: I updated the code as suggested

package.json
```diff
@@ -35,6 +35,7 @@
     "wds": "wds --app-index dev/index.html --watch --node-resolve",
     "preserve": "yarn prep",
     "serve": "concurrently -c \"cyan,blue\" \"yarn:rollup.watch\" \"yarn:wds\" --kill-others",
+    "storybook.axe": "yarn storybook.prod && yarn axe-storybook --build-dir=storybook-static/",
```
damontgomery, line 38: What about `test.accessibility` as the name. The rest of the storybook commands are about running storybook. This command, to me falls under the testing category.

damontgomery (approved):
This looks good! I had tried `storybook-addon-storypa11y` and couldn't get it to work.

Added a suggestion for a command name.

In a chat, we discussed ignores.

If we figure that out, it would be good to know if we need to ignore a different way for this tool and for the AXE tools that is part of the addons in the Storybook UI. Maybe create a story for that?

I've used that before:
```javascript
parameters: {
  a11y: {
    config: {
      rules: [
        {
          // The contrast isn't being accurately measured.
          id: 'color-contrast',
          enabled: false,
        },
      ],
    },
  },
},
```

package.json
shaal: Interesting. This accessibility test runs against storybook stories.

himerus: @damontgomery you may be correct. And we may want a better naming scheme for some of these things (all of them) as we move forward. 

For instance in our `fin` scripts, we stopped using the names of a tool specifically (like Storybook), and abstracted the command names to be about the result. For now we'll go with this, but I know we have a ticket somewhere to "review" the `package.json` for ease of use and all that good stuff.

himerus (approved):

PR #235, opened 2022-02-25
fix: switch prop from lang to language

grshane:


shaal (approved):
Thank you! 
This fixes one of the accessibility issues in our components.

himerus (approved):

Issue #236, opened 2022-02-28
Add Header Mobile Updates


sarahp:
Add mobile styles and the mobile controller to `outline-heading`. As shown on ghs: https://github.com/phase2/ghs-design/blob/next/src/components/base/outline-heading/outline-heading.ts#L56.

PR #237, opened 2022-02-28
#236: Adding in mobile controller and CSS config changes to Outline Heading

sarahp:
**Developer Notes**
Adding in the `mobile controller` and CSS config changes into Outline Heading. Implements Issue #236  

**Testing Notes**
1. Go to Storybook
2. Go to the Heading component
3. Size the screen up and down
4. Observe that the text gets smaller at smaller screens and larger at larger screens

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/237"></a>


himerus:
Holding this one until we see the full impact and any issues with these changes on an internal client project. I think we may have a few small adjustments to make here.

shaal:
@sarahp I'd love to understand this feature better.
Aren't heading display sizes very specific in each project?
What does the mobile controller do?

sarahp:
> @sarahp I'd love to understand this feature better. Aren't heading display sizes very specific in each project? What does the mobile controller do?

The mobile controller applies where to break. This doesn't change the sizes of the `h1-h6` or the `levels`, just sets the sizes to use a smaller `level` or `h-tag` when the screen gets smaller.

himerus:
Updated with various cleanup, better usage of all 3 sizes on `<outline-heading>` elements that only use the `level` attribute and omit the `level-size` one.

PR #238, opened 2022-03-01
yarn upgrade + rollup-plugin-typescript2

shaal:
Solving an issue (on Gitpod, node v17), when running `yarn build`

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/238"></a>


himerus (approved):
As long as the pipelines pass, I'm good with this. The same test seemed solid in local environments as well.

PR #240, opened 2022-03-03
Button Upgrades

himerus:
image

image

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/240"></a>

### DIFF
--- a/src/components/base/outline-button/outline-button.stories.ts
+++ b/src/components/base/outline-button/outline-button.stories.ts
@@ -1,19 +1,15 @@
 import { html, TemplateResult } from 'lit';
 import { ifDefined } from 'lit/directives/if-defined.js';
-// import { AllIcons } from '../outline-icon/config';
 import { argTypeTarget } from '../outline-link/config';
-import { argTypeSlotContent } from '../../base/outline-element/utils/utils';
-import { ButtonVariant, ButtonSize } from './outline-button';
+import {
+  argTypeHidden,
+  argTypeSlotContent,
+} from '../outline-element/utils/utils';
+import { ButtonVariant } from './outline-button';
 import './outline-button';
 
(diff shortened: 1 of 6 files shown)


src/components/base/outline-button/outline-button.stories.ts
```diff
@@ -1,62 +1,73 @@
...
 export default {
   title: 'Navigation/Button',
   component: 'outline-button',
   argTypes: {
     ...argTypeSlotContent,
-    target: argTypeTarget,
+    target: {
+      name: 'button-target',
+      description: 'https://www.w3schools.com/tags/att_a_target.asp',
+      options: ['_blank', '_self', '_parent', '_top'],
```
shaal: @himerus We are missing the option `(none)`
(also in `outline-link`)

shaal (changes requested):
LGTM, 1 comment about adding (none) option in storybook, for attributes that are optional.

src/components/base/outline-button/outline-button.stories.ts
himerus: @shaal So you're just wanting/expecting the ability to unselect/remove the target attribute in the Story?

shaal: Exactly, but this could be a separate PR

himerus: Got it in. Also defaulted to a REAL url, and `_blank` for the target, though you CAN deselect it to change that. I wanted it to be blank page by default just because everyone is going to click that darn button/link.

shaal (approved):
LGTM, thank you!

src/components/base/outline-button/outline-button.stories.ts
shaal: 👍

PR #262, opened 2022-03-09
Feature: Bubble events

damontgomery:
## Description

Add a way to bubble / redispatch events up the DOM tree so that things like Google Tag Manager can watch for them.

For now, included `submit` as a default option.

The new custom events provide
- The same `event.type` name
- An attempt to provide getters for all the properties on the source event
- `event.sourceEvent` which is the original event and can take functions like `preventDefault()`
- `event.aggregatedComposedPath()` which is a substitute for `event.composedPath()` that attempts to show the full path through the DOM

## Type of change

## How Has This Been Tested?

- Visit http://localhost:6042/?path=/story/tests-bubble-events-submit--submit
- Enable `Preserve log` in Chrome inspector (see below)
- Click on the submit button
- View the logged event

Enable_preserve_log

story

log_messages

## Checklist:

### Testing
I'm not sure if there is a better way to provide a test for this, so happy to hear other suggestions.

I know some projects (including the one that needs something like this) have an `x-form` component and if we get that into Outline, the test story / component could be deleted.

One of the key parts of testing was testing nested shadow DOMs (not slots). Wanted to make sure that the events can reach beyond a single shadow DOM and that we also only get one event per context. The provided test allowed me to check those conditions.

### Context

This is done as work for a client project, but it will hopefully resolve an issue for all projects that use Outline and we can easily define additional event types as needed.

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/262"></a>

### Attempted alternative ways to clone events

I tried `Proxy` objects, but I couldn't find a way to dispatch those.

I tried to clone the object with `new event.constructor(event.type, event);`, but I could not set the `bubbles` or `composed` properties which are necessary so that the event continues to bubble through multiple shadow DOMs.

I tried `Object.defineProperty(eventForLightDOM, 'composed', true)`, but that didn't seem to actually make the event bubble.

There may be a better way. Please let me know or try it out with the provided test story.


damontgomery:
@himerus & @grshane, I noticed you both worked with `Rollup` issues / differences with imports. I'm having a hard time getting it to work where I can import the object and have both `yarn rollup` and `yarn start` happy.

With the current state, at least they don't fail, but Rollup gives an error and not sure if the distributed version works. The storybook version does work.

Appreciate any help, but this is not urgent.

...

Yeah, this isn't working. The `yarn serve` example page gives this issue: `Uncaught ReferenceError: module is not defined
    at outline.config.js:6:1`. That's the place where we try to export the object.

Is there some way we can use another type of config file? This format seems like ES5.

himerus:
Yeah, let's chat. I totally have a fix for this. It should already be built into this latest version. :) 
Checkout after running `yarn prep` the new `src/resolved-outline-config.ts`. This is the version you want to import "most of the time" in our component or story files.

The one at the root can be included/used (without complaints) in the Node scripts in the `scripts` directory.

himerus:
Additionally, the `outline-form` element is a VERY soon thing, and I expect us to TRY to use it on all projects until we find something that breaks everything and have to fight the big fires.

damontgomery:
Doing more testing on this to determine limitations.

damontgomery:
Fixed an issue with duplicate cloned events. Testing still WIP on internal project, but looking promising. :)

damontgomery:
I need to test this more and make sure multiple nested shadow DOMs work as expected. I'm seeing the `composed` property set to `false` which worries me.

PR #266, opened 2022-03-23
Fixes for `slotController`

shaal:
## Description

When using partial template, and displaying it in different location in the DOM, according to mobile/desktop - slots that are moved into shadow DOM, disappear from the display. 

Fixes # (issue)

Instead of moving slots into Shadow DOM, cloning them works much better.
* The original slot stays in the lightDOM (it never gets displayed)
* `shadowSlotLocation` consist the original link ("reveal" when inspecting elements), that usually display the slot in the correct location.

This PR:
1. Clones the slotted content into the right location (before `shadowSlotLocation`)
1. Removes `shadowSlotLocation` so the original link of light DOM content is no longer there.
1. Each slot type has to be cloned in a special way:
    1. named-slot
    1. default-slot--content
    1. default-slot--element

Breaking changes:
To style a specific cloned slot, write in CSS:
```css
div[cloned-slot-name='background-image'] {
  height: 100%;
}
```

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/266"></a>


himerus (changes requested):
This still needs work. 
I see many original non-edge cases of the slotController now broken in current project. I would suspect, based on the components I'm seeing it on that this is in relation to double or triple nested ShadowDOMs. 

Will do more investigation in an hour or two, but for the moment, this should NOT be merged.

src/components/controllers/slot-controller.ts
```diff
@@ -132,59 +134,80 @@ export class SlotController implements ReactiveController {
...
-      const slotLightDom = host.querySelector('[slot=' + slotName + ']');
-
-      // Only if named slot found in light dom - move it into ShadowDOM.
-      if (slotLightDom) {
-        shadowSlotLocation.append(slotLightDom);
-      }
-    }
+  private moveNamedSlots(slotShadowDom: HTMLSlotElement) {
+    const slotLightDomArray = this.hostEl.querySelectorAll(
+      `:scope > [slot="${slotShadowDom.name}"]`
```
danmuzyka: It looks as though this line limits this logic to slots that are a direct child of the host element. So the implication would be that this would not apply to a slot wrapped in a `<div>`, for example. Is that the intention, or am I misunderstanding something here?

damontgomery: The idea is to avoid affecting slots from child web components. See https://github.com/phase2/outline/pull/304 for a demonstration. If you have nested components with the same slot name, you could get elements from the wrong element when you do this query without `:scope >`.

`this.hostEl` is the components light DOM. If you have slots with web components in them, they should handle their own behavior.

I could be wrong, but I believe for slots to behave as expected, they need to be direct children of the light DOM component. Is there a use case for wrapping slots in other elements and having them function as slots on the component?

shaal: I think I'll replace that selector with `assignedSlot`, which seems to me "cleaner" way of using the browser API, but will result in EXACTLY the same thing.

`moveNamedSlots` is receiving the slot in the shadow Dom, which is a placeholder/pointer to the slot in Light DOM.
So there's no limitation at all here, because when a slot is wrapped with `<div>`, that div is already in the shadow DOM.

danmuzyka: OK, I think I understand how this works, now. To paraphrase, the selector is actually pointing to the slotted Light DOM node, which is always a direct child of the host element, irrespective of where it then gets dropped into the Shadow DOM tree.

Thanks for the detailed explanations, @damontgomery and @shaal !!

damontgomery: Yes, I believe you've got it. I get mixed up with all these DOMs, so no worries. :)

himerus (changes requested):
Tests need updating.

shaal:
> Tests need updating.

Done 🎉

PR #267, opened 2022-03-24
`outline-google-map` Component

peterschuelke:
## Description

This creates a new Google Map component for Outline.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/267"></a>

### DIFF
--- a/src/components/base/outline-google-map/outline-google-map.stories.ts
+++ b/src/components/base/outline-google-map/outline-google-map.stories.ts
@@ -1,0 +1,202 @@
+import { html, TemplateResult } from 'lit-html';
+import { ifDefined } from 'lit/directives/if-defined.js';
+import { unsafeHTML } from 'lit/directives/unsafe-html.js';
+import { argTypeSlotContent } from '../../base/outline-element/utils/utils';
+import './outline-google-map';
+import './src/outline-map-marker';
+
+import { MapType, GestureType } from './outline-google-map';
+const mapTypeOptions: MapType[] = ['roadmap', 'satellite', 'hybrid', 'terrain'];
+const gestureOptions: GestureType[] = ['auto', 'cooperative', 'greedy', 'none'];
+
+export default {
+  title: 'Content/Map',
+  component: 'outline-google-map',
+  argTypes: {
+    ...argTypeSlotContent,
+    apiKey: {
+      description: '**REQUIRED:** Sets the Google Maps api key.',
+    },
+    centerLatitude: {
+      description:
+        '**OPTIONAL Number:** Sets the starting map lattitude. FitToMarkers overrides.',
+    },
+    centerLongitude: {
+      description:
+        '**OPTIONAL Number:** Sets the starting map longitude. FitToMarkers overrides.',
+    },
+    zoom: {
+      description:
+        '**OPTIONAL Number:** Sets the starting map zoom level. FitToMarkers overrides on inital page load, adjusting in Storybook will override FitToMarkers.',
+    },
+    fitToMarkers: {
+      description:
+        '**Boolean**: Sets the starting map zoom level to show all markers. Overrides zoom and center.',
+    },
+    mapType: {
+      description:
+        '**OPTIONAL MapType** Sets the style of map ("roadmap", "satellite", "hybrid", "terrain").',
+      options: mapTypeOptions,
+      control: {
+        type: 'select',
+      },
+    },
+    gestureHandling: {
+      description:
+        '**GestureType** Sets how the map handles touch events ("auto", "cooperative", "greedy", "none").',
+      options: gestureOptions,
+      control: {
+        type: 'select',
+      },
+    },
+    disableDefaultUI: {
+      description: '**Boolean**: Clears the map controls.',
+    },
+    zoomControl: {
+      description:
+        '**Boolean**: Individually adds zoom controls to map after disabling controls.',
+    },
+    mapTypeControl: {
+      description:
+        '**Boolean**: Individually adds map type controls to map after disabling controls.',
+    },
+    scaleControl: {
+      description:
+        '**Boolean**: Individually adds scale controls to map after disabling controls.',
+    },
+    streetViewControl: {
+      description:
+        '**Boolean**: Individually adds street view controls to map after disabling controls.',
+    },
+    rotateControl: {
+      description:
+        '**Boolean**: Individually adds rotation controls to map after disabling controls.',
+    },
+    fullScreenControl: {
+      description:
+        '**Boolean**: Individually adds full screen controls to map after disabling controls.',
+    },
+  },
+  parameters: {
+    viewMode: 'docs',
+    previewTabs: {
+      canvas: {
+        hidden: false,
+      },
+    },
+    docs: {
+      description: {
+        component: `This component renders a Google Map.`,
+      },
+      source: {
+        code: `
+<outline-google-map
+  api-key="{{ apiKey }}"
+  map-type="{{ mapTypeId }}"
+  center-latitude="{{ mapStartingLat }}"
+  center-longitude="{{ mapStartingLong }}"
+  fit-to-markers
+  disable-ui
+>
+  <outline-map-marker
+    slot="markers" 
+    latitude="{{ locationLat }}" 
+    longitude="{{ locationLong }}"
+    icon="{{ pngUrlorOutlineSVG }}"
+    iconFillColor="{{ htmlColororOutlineVar }}">
+      {{ mapMarkerDefaultSlot }}
+  </outline-map-marker>
+</outline-google-map>
+        `,
+      },
+    },
+  },
+};
+
+const Template = ({
+  apiKey,
+  zoom,
+  mapType,
+  centerLatitude,
+  centerLongitude,
+  fitToMarkers,
+  disableDefaultUI,
+  zoomControl,
+  mapTypeControl,
+  scaleControl,
+  streetViewControl,
+  rotateControl,
+  fullScreenControl,
+  gestureHandling,
+  defaultSlot,
+}): TemplateResult => {
+  zoom = zoom > 0 ? zoom : null;
+  return html`
+      <outline-google-map
+        apiKey="${apiKey}"
+        .zoom="${zoom}"
+        mapType="${mapType}"
+        centerLatitude="${ifDefined(centerLatitude)}"
+        centerLongitude="${ifDefined(centerLongitude)}"
+        gestureHandling="${gestureHandling}"
+        ?fitToMarkers="${fitToMarkers}"
+        ?disableDefaultUI="${disableDefaultUI}"
+        ?zoomControl="${zoomControl}"
+        ?mapTypeControl="${mapTypeControl}"
+        ?scaleControl="${scaleControl}"
+        ?streetViewControl="${streetViewControl}"
+        ?rotateControl="${rotateControl}"
+        ?fullScreenControl="${fullScreenControl}"
+      >
+      ${unsafeHTML(`${defaultSlot}`)}
+      </outline-google-map>
+    `};
+
+export const OutlineGoogleMap = Template.bind({});
+OutlineGoogleMap.args = {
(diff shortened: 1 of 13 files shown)


src/components/base/outline-google-map/outline-google-map.ts
```diff
@@ -0,0 +1,248 @@
+import { LitElement, html, CSSResultGroup } from 'lit';
+import { customElement, property, query } from 'lit/decorators.js';
+import { GoogleMapsApi } from './src/google-maps-api';
+import './src/google-maps-api';
+import { OutlineMapMarker } from './src/outline-map-marker';
+import { MapSelector } from './src/map-selector';
+import './src/map-selector';
```
damontgomery, line 7: Minor question:

Are there ways to combine these imports? A reason to keep them separate? I'm still trying to understand best practices for imports and the trade-offs.

src/components/base/outline-google-map/outline-google-map.ts
```diff
@@ -0,0 +1,248 @@
...
+import componentStyles from './outline-google-map.css.lit';
+
+export type MapType = 'roadmap' | 'hybrid' | 'satellite' | 'terrain';
+export type GestureType = 'auto' | 'cooperative' | 'greedy' | 'none';
```
damontgomery, line 11: Minor suggestion:

If you use the pattern below you can import the list of options in the story so you only need to define it once.
```javascript
export const buttonSizes = ['small', 'medium', 'large'] as const;
export type ButtonSize = typeof buttonSizes[number];
```

src/components/base/outline-google-map/outline-google-map.ts
```diff
@@ -0,0 +1,248 @@
...
+
+@customElement('outline-google-map')
+export class OutlineGoogleMap extends LitElement {
+  static styles: CSSResultGroup = [componentStyles];
+  /**
+   * A Maps API key. To obtain an API key, see https://developers.google.com/maps/documentation/javascript/tutorial#api_key.
+   */
+  @property({ type: String })
+  apiKey = '';
```
damontgomery, line 20: Minor: We all have different opinions, but I recommend always using the options below:

`{ type: ..., reflect: true, attribute: ... }`

so

`{ type: String, reflect: true, attribute: 'api-key' }`

The reasoning:
- Consistent kabob case for attributes
- Reflection makes the developer inspection match the component state if you modify it in some way. Otherwise you can end up in situations where someone looks at the inspector and the state and attributes have diverged and that seems unnecessarily confusing.

src/components/base/outline-google-map/outline-google-map.ts
```diff
@@ -0,0 +1,248 @@
...
+   */
+  @property({ type: String })
+  version = '3.48';
+
+  /**
+   * If set, custom styles can be applied to the map.
+   * For style documentation see https://developers.google.com/maps/documentation/javascript/reference#MapTypeStyle
+   */
+  @property({ type: Object })
+  styles: google.maps.MapTypeStyle[] = [];
```
damontgomery, line 33: How does this work with an HTML consumer? Does this only work with a JS framework?

When we've extended styles, we've extended the component, but there may be better ways.

There is also a way to process attributes and convert them to properties that might work if you want to pass in JSON. If it doesn't do that already.

src/components/base/outline-google-map/outline-google-map.ts
```diff
@@ -0,0 +1,248 @@
...
+  language = '';
+
+  @property({ type: String })
+  mapId = '';
+
+  map: google.maps.Map;
+
+  markers: Array<OutlineMapMarker>;
+
+  marketObserverSet: boolean;
```
damontgomery, line 99: Minor: Is this a typo? `market` instead of `marker`?

src/components/base/outline-google-map/outline-google-map.stories.ts
```diff
@@ -0,0 +1,202 @@
...
+  disableDefaultUI,
+  zoomControl,
+  mapTypeControl,
+  scaleControl,
+  streetViewControl,
+  rotateControl,
+  fullScreenControl,
+  gestureHandling,
+  defaultSlot,
+}): TemplateResult => {
```
damontgomery, line 132: Minor:

I usually use this pattern for stories since listing all the variables just means I have to track them in more than one place.

```javascript
const Template = (args = configuration.args): TemplateResult => {
  args = {
    ...configuration.args,
    ...args,
  };

  return html``;
};
```

Then you can access `args.myAttribute`.

src/components/base/outline-google-map/outline-google-map.stories.ts
```diff
@@ -0,0 +1,202 @@
...
+  zoom = zoom > 0 ? zoom : null;
+  return html`
+      <outline-google-map
+        apiKey="${apiKey}"
+        .zoom="${zoom}"
```
damontgomery, line 137: Any reason to set a property here rather than the attribute `zoom`?

src/components/base/outline-google-map/outline-google-map.stories.ts
```diff
@@ -0,0 +1,202 @@
...
+        ?streetViewControl="${streetViewControl}"
+        ?rotateControl="${rotateControl}"
+        ?fullScreenControl="${fullScreenControl}"
+      >
+      ${unsafeHTML(`${defaultSlot}`)}
+      </outline-google-map>
+    `};
+
+export const OutlineGoogleMap = Template.bind({});
+OutlineGoogleMap.args = {
```
damontgomery, line 156: Minor: I usually define defaults in the configuration `args` property. That way if we need to make another story later, we don't need to redefine everything, just the changes.

src/components/base/outline-google-map/src/google-maps-api.ts
```diff
@@ -0,0 +1,140 @@
...
+  connectedCallback() {
+    super.connectedCallback();
+    this.isReady = true;
+    if (this.libraryUrl != null) this.loadLibrary();
+  }
+}
+// TODO: Resolve ts error: This syntax requires an imported helper but module 'tslib' cannot be found.
+// eslint-disable-next-line
+// @ts-ignore
+@customElement('google-maps-api')
```
damontgomery, line 59: Question: Is there a benefit of having this be a Lit element rather than something else?

It doesn't appear that anything is rendered here, so maybe this could just be a general module that exports functionality for loading the library?

src/components/base/outline-google-map/src/jsonp-library.ts
```diff
@@ -0,0 +1,54 @@
+import { LitElement } from 'lit';
+import { ScriptLoaderMap } from './script-loader-map';
+
+export abstract class JsonpLibraryElement extends LitElement {
```
damontgomery, line 4: Question: Same as for `google-maps-api.ts` Any reason this is a LitElement?

src/components/base/outline-google-map/src/map-selector.ts
```diff
@@ -0,0 +1,137 @@
...
+  _items: Array<Node> = [];
+
+  get items(): Array<Node> {
+    return this._items;
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    this.addEventListener('slotchange', event => {
```
damontgomery, line 29: Does this work? The `slotchange` occurs on slots, not components or slotted elements. So, I don't think this will ever fire.

I may be wrong.

If you want to watch for changes, I think you need to add this to a slot or add a mutation observer to the element.

src/components/base/outline-google-map/outline-google-map.stories.ts
```diff
@@ -0,0 +1,202 @@
...
+
+import { MapType, GestureType } from './outline-google-map';
+const mapTypeOptions: MapType[] = ['roadmap', 'satellite', 'hybrid', 'terrain'];
+const gestureOptions: GestureType[] = ['auto', 'cooperative', 'greedy', 'none'];
+
+export default {
+  title: 'Content/Map',
+  component: 'outline-google-map',
+  argTypes: {
+    ...argTypeSlotContent,
```
damontgomery, line 16: Can we create a Storybook argType / arg for the `markers` slot? Right now it shows as `Default Slot` which seems misleading.

damontgomery (commented):
Thanks @peterschuelke!

I left a bunch of questions / suggestions / comments, but I want to start with...

THIS IS VERY COOL!

The component works well and the Storybook configuration is very cool.

Overall, I think the code was pretty clear on a function by function level.

Could we add some documentation around how all the parts fit together? I think this may be helpful for someone less familiar with Google Maps. Maybe a `readme.md`? This seems like one of our most complex components by far. At least, that I've seen.

My main question about the architecture is around the modules in `src`. Is there any benefit to having these be `LitElement` modules? I don't think they render anything and maybe they could just be functions / classes / etc. I've never seem `LitElement` used this way, but that's not to say it isn't a good approach.

Thanks again and hope some of my comments / suggestions / questions are helpful.

src/components/base/outline-google-map/outline-google-map.ts
damontgomery: https://lit.dev/docs/components/properties/#conversion-converter

src/components/base/outline-google-map/outline-google-map.ts
peterschuelke: The code works with just importing the entire file, but tsc complains about using the classes OutlineMapMarker and MapSelector without importing it specifically.

$yarn tsc
yarn run v1.22.10

$ /Users/peterschuelke/sites/app.ochsner.org/design/node_modules/.bin/tsc

src/components/base/outline-google-map/outline-google-map.ts:107:13 - error TS2304: Cannot find name 'MapSelector'.

107   _markers: MapSelector;

                ~~~~~~~~~~~

src/components/base/outline-google-map/outline-google-map.ts:107:13 - error TS4031: Public property '_markers' of exported class has or is using private name 'MapSelector'.

107   _markers: MapSelector;

                ~~~~~~~~~~~

Found 2 errors.

error Command failed with exit code 1.

info Visit https://yarnpkg.com/en/docs/cli/run  for documentation about this command.

damontgomery: Thanks for the context. :)

src/components/base/outline-google-map/outline-google-map.ts
peterschuelke: I took a stab at this and I'm not sure it's properly enforcing the type. I use:
`export const gestureTypes = ['auto', 'cooperative', 'greedy', 'none'];
export type GestureType = typeof gestureTypes[number];`

I can then set without errors:
`  @property({ type: String })
  gestureHandling: GestureType = 'green';`

src/components/base/outline-google-map/src/google-maps-api.ts
peterschuelke: By setting `GoogleMapsApi` as an extension of `JsonpLibraryElement` which itself is an extension of `LitElement` we are getting the lifecycle benefits of a LitElement. We are also wanting to prevent the google maps library to load multiple times. Having this as a component we run the risk of SPA and multiple component use causing memory leaks. This is a method of mitigating those memory leaks. I'm sure there's other ways to achieve these objectives as well.

src/components/base/outline-google-map/outline-google-map.ts
damontgomery: I think you need this, with `const` at the end.

```javascript
export gestureTypes = ['auto', 'cooperative', 'greedy', 'none'] as const;
```

I don't remember the specifics, but I think what you have tells it that you have an array of strings, but this says the array is a constant and so it has to be one of those values.

damontgomery: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

src/components/base/outline-google-map/src/google-maps-api.ts
damontgomery: Thanks for the information. :)

PR #269, opened 2022-03-30
`outline-button-group` Component

### LINKED ISSUE #44: Button Group
## Requirements

Create a basic button group component that:

- Takes pre-rendered instances of `outline-button` passed into a slot.
- Has basic 'Horizontal', 'Vertical' styling options.
- Meets accessibility standards.

### Example
https://designsystem.digital.gov/components/button-group/
Button_group___U_S__Web_Design_System__USWDS_

### Accessibility Guidelines
[a11y Style Guide](https://a11y-style-guide.com/style-guide/section-general.html#kssref-general-buttons) (also look at Tabs info?)
[Digital A11Y group role](https://www.digitala11y.com/group-role/)

### Notes/Questions
Not sure if we need to have some kind of state/change logic built into this component.
This should be actually simple to just pass in any number of buttons into a slot, and then based on a single property of horizontal/vertical alignment, you can swap CSS.

rbeach:
## Description

Adds 

Resolves #44: https://github.com/phase2/outline/issues/44

## Type of change

[X] New feature (non-breaking change which adds functionality)

This branch adds `outline-button-group`, a component that essentially wraps `outline-list` and provides sensible defaults for using the list component to render a group of buttons.  One would assume that a button group would not use `ol`, so this component forces the use of ul.

Would it be better than this current wrapper implementation to have the `outline-button-group` component extend the `outline-list-component`? From my understanding, "extending" a component would offer _more_ options, but here we want to limit the options that `outline-list` gives us.

## How Has This Been Tested?

Tests copy the boilerplate assertions from other components

- [X] Renders with Default Values
- [ ] Renders with Slotted content
I couldn't seem to get this working. I would assume it would be written something like:

```
it('renders slotted content', async () => {
    const el = await fixture(
      html`<outline-button-group>
        <li>
          <button>test button</button>
        </li>
      </outline-button-group>`
    );
    assert.lightDom.equal(
      el,
      `
      <ul>
        <li>
          <button>
            test button
          </button>
        </li>
      </ul>
      `
    );
  });
```

I wonder if, because this component is essentially set up as a wrapper for `outline-list`, and the slot is coming from that component, if there is a different way to write this.

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [-] I have added tests that prove my fix is effective or that my feature works
- [X] Any dependent changes have been merged and published in downstream modules

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/269"></a>


himerus:
@rbeach My first instinct here, is that though the `outline-list` component provides the horizontal/vertical alignments wanted here, we don't need the complexity of that in the background and don't really need the buttons in a ul/li structure. I will have @shaal chime in if there's a different thought there for a11y. 

So I'd just extend `OutlineElement`, and add in the horizontal/vertical features from the `outline-list`. I'd expect the markup in the story to just be a bunch of buttons dropped in a group. I may be wrong. Going to ping for some other input.

himerus:
@rbeach I think we can figure out something here in order to slim down the options available. Typically, we've ADDED new props/features to an extended components, but I'm trying to think of an example to NOT use them. 

I'll look again when I have a chance. It might be as simple as rewriting the `render` method to exclude any logic implied by those properties we don't want. Since we really ONLY want the orientation.

PR #285, opened 2022-04-06
feat: added linked-block controller

madhaze:
## Description

This MR adds link-block-controller

Fixes # (issue)

## Type of change

- [x] New feature (non-breaking change which adds functionality)

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/285"></a>

### DIFF
--- a/src/components/controllers/linked-block-controller.ts
+++ b/src/components/controllers/linked-block-controller.ts
@@ -1,0 +1,106 @@
+import { ReactiveController } from 'lit';
+import { OutlineElement } from '../base/outline-element/outline-element';
+
+interface OutlineElementWithHref extends OutlineElement {
+  href?: string;
+}
+
+/**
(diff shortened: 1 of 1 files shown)


src/components/controllers/linked-block-controller.ts
```diff
@@ -0,0 +1,106 @@
...
+  hostConnected() {
+    if (this.isLinked()) {
+      this._handleClicks();
+      this._handleHover();
+      this._handleFocus();
+    }
+  }
+
+  isLinked(): boolean {
+    return this.host.href !== undefined && this.host.href.length > 0;
```
damontgomery, line 51: Want to point out that this uses the pattern, if component has `href` as an attribute, it will function as a link.

I believe we decided we didn't want to use that pattern of using attribute names that already exist.

I think @sarahp updated this for a project.

PR #291, opened 2022-04-08
Gitpod fix: use node-lts as docker image (node v16)

shaal:
## Description

Instead of using `nvm` in Gitpod, we should use node-lts as a docker image, which is node v16.

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/291"></a>

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -4,6 +4,8 @@ image:
 tasks:
   - init: |
       yarn
+      nvm install $(cat .nvmrc)
+      nvm use
     command: |
       yarn start
 
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -4,8 +4,6 @@ image:
 tasks:
   - init: |
       yarn
-      nvm install $(cat .nvmrc)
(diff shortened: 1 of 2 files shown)


.gitpod.yml
```diff
@@ -4,8 +4,6 @@ image:
 tasks:
   - init: |
       yarn
-      nvm install $(cat .nvmrc)
-      nvm use
```
himerus: Just because we are using a LTS version of the Docker image doesn't necessarily mean we will want to switch to it the second it is released. It would required testing prior to deciding to switch. Eventually we'll need/want to run tests on multiple node versions at the same time in our pipelines. That will help upgrading to a new LTS version easier. 

We include the `.nvmrc` in order to restrict (or recommend) the desired version of Node. 

Yes, let's use the LTS version of the image, but let's not remove this section.

himerus (changes requested):

.gitpod.yml
shaal: 👍

PR #328, opened 2022-06-01
fix: Update outline-heading margin bottom styles.

jardix22:
## Description

The margin-bottom was moved from the `headings` tag selector to the `:host` selector to allow it to be overridden. Currently, with the margin defined on the `heading` tags selectors (see attached code), it is not possible to override it due to the shadow DOM protection.

```CSS
/* file: src/components/base/outline-heading/outline-heading.css */

h1,
h2,
h3,
h4,
h5,
h6 {
  @apply m-0 mb-4;
}
```

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [x] Visual Testing

  Heading component with margin-bottom styles on headings tag selector
 
  image

  Heading component with margin-bottom styles on the `:host` selector.
 
  image

As you can see in the attached images of above this change does not create any regression.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/328"></a>

### DIFF
--- a/src/components/base/outline-heading/outline-heading.css
+++ b/src/components/base/outline-heading/outline-heading.css
@@ -1,5 +1,5 @@
 :host {
-  @apply block;
+  @apply mb-4 block;
 }
 
 h1,
@@ -8,7 +8,7 @@ h3,
 h4,
 h5,
 h6 {
-  @apply m-0 mb-4;
+  @apply m-0;
 }
 
 /* Default sizes, can be changed with level-size prop */


shaal:
@jardix22 what is the use case of overriding a component style?
It's in the shadowDom in order for nothing else to modify it, no?

My personal take - Until we use Utopia for fluid spacing, I like heading to have margin 0, and add margins according to design.

If we want to override styles in shadowDom, we can use `part:`, or we could use css variables.

jardix22:
_What is the use case of overriding a component style?_
I saw some components where they create their own heading styles instead of using the `outline-heading` due to the margin-bottom override limitation. 

---

@shaal You're right, I didn't know about the `::part` pseudo-element, which solves the overring limitation.

```
<${unsafeStatic(this.level as string)} part="heading" class=${classMap(classes)}>
  <slot></slot>
</${unsafeStatic(this.level as string)}>
```

So, would you agree with the `part` attribute addition on the `outline-heading` markup?

Thank you for your comments!

PR #330, opened 2022-06-02
Issue/329 mobile controller update

### LINKED ISSUE #329: [Bug]: MobileController: Fix type error and consider renaming to BreakpointController
**Describe the bug**

- Resolve type error here: https://github.com/phase2/outline/blob/next/src/components/controllers/mobile-controller.ts#L26
- Consider renaming the controller to BreakpointController

mabry1985:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #329

## Type of change

- [x] Breaking change * This would affect any projects using the mobile controller in their design system if they wanted to update to the latest Outline

## How Has This Been Tested?

Verified that the `outline-accordion`, `outline-tab-group`, `outline-dropdown`, and `outline-heading` which were implementing mobile controller are still functioning as expected

- [x] Visual Testing

## Checklist

<a href="https://gitpod.io/#https://github.com/phase2/outline/pull/330"></a>

### DIFF
--- a/src/components/base/outline-tabs/outline-tab-group/outline-tab-group.ts
+++ b/src/components/base/outline-tabs/outline-tab-group/outline-tab-group.ts
@@ -60,7 +60,7 @@ export default class OutlineTabGroup extends OutlineElement {
     window.addEventListener('resize', this._handleResize);
 
     this.resizeObserver = new ResizeObserver(() => {
-      if (!this.breakpointController.isMobile) {
+      if (!this.breakpointController.isBreakpoint) {
         this.preventIndicatorTransition();
         this.repositionIndicator();
       }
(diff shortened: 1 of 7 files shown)


src/components/base/outline-tabs/outline-tab-group/outline-tab-group.ts
```diff
@@ -60,7 +60,7 @@ export default class OutlineTabGroup extends OutlineElement {
     window.addEventListener('resize', this._handleResize);
```
shaal, line 60: We should not use addEventListener with `resize` (old method), it's very costly for performance.
That's what `ResizeObserver` is for (new method)

shaal (commented):
All `.isMobile` instances should be updated to `.isBreakpoint`

shaal:
Is the `screen-size-controller` already doing what `breakpoint-controller` supposed to do?
https://github.com/phase2/outline/blob/next/src/components/controllers/screen-size-controller.ts

mabry1985:
@shaal good catch on the `.isMobile` I never saw that flagged locally.

As far as updating to use the resizeObserver method or sunsetting this controller, I think thats worth discussing. We can decide how we want to proceed and update issue #329 or close it out.

PR #380, opened 2023-04-29
Add docs and improve developer experience.

tekante:
## Description

Added ability for project initialization to use local template files and documentation to help guide those developing for Outline.

## Type of change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How Has This Been Tested?

- [x] Manual testing locally

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #384, opened 2023-04-30
Local development enhancements

tekante:
## Description

Further enhancements for local development and addition of successful project creation via ability to run `yarn build` and `yarn storybook:build` within the project. This also has changes to test for successful project initiation in CI. See https://github.com/phase2/outline/actions/runs/4846186198 for example of a full successful run.

This builds on work from branch `feature/docs-and-init-dx` so I'm setting that as the target at the moment to make it easier to review what has changed since then.

Once #380 merges this can be targeted at the `next` branch for merging though note that the changes in #385 are needed for full successful CI.

## Type of change

## How Has This Been Tested?

- [ ] Visual Testing
- [ ] Automated Testing
- [ ] Accessibility Testing

## Checklist


.github/workflows/build.yml
```diff
@@ -123,6 +118,54 @@ jobs:
       - name: Build Storybook
         run: yarn storybook:build
 
+  # Test initialization of a new project
+  build_project:
+    name: Initialize and Test Build of Project Codebase
+    runs-on: ubuntu-latest
```
mike-potter, line 123: Are we going to run out of Github action minutes using their runner here instead of `self-hosted`?

bardleb (approved):

shaal (approved):
@tekante thank you!

packages/outline-templates/default/scripts/link-for-local-dev.js
```diff
@@ -0,0 +1,81 @@
+/**
+ * Takes an argument of the path to a directory containing package directories to link for
+ * local development. Will register them all for local development in the current project.
+ */
```
himerus, line 4: Now that I'm looking through things, I'm thinking running `yarn turbo dev` does this essential same linking of the local packages. 

I've not used the mode but @grshane reminded me it was a thing.

himerus: So I'm pondering if we can be running `yarn turbo dev` and then run the CLI commands in a similar way.. .

himerus (commented):
I think we may have some other options here. I've really got to dive further into the details here and what turborepo should be doing for us by default.

.github/workflows/build.yml
tekante: I've confirmed that builds from this repo aren't counting against our action minutes allocation.

packages/outline-templates/default/scripts/link-for-local-dev.js
tekante: This script is intended for use in a generated project. I don't think `yarn turbo dev` would apply in that context. Am I overlooking something?

himerus: Derp. In my head this was for Outline local dev.

himerus: This is what I was thinking of where `yarn turbo run dev` would possibly serve the same purpose as what's going on there. https://github.com/phase2/outline/actions/runs/5046433815/jobs/9052038359#step:7:15

It's worth some digging later.

PR #385, opened 2023-04-30
Feature/comprehensive updates

tekante:
## Description

Changes to allow for resulting initiated project to successfully run `yarn run start` and `yarn run build` PR'd against branch allowing for CI testing of project initiation so the changes can be annotated pointing at the errors that required them.

This should probably target the `next` branch and merge along with #384. See files changed for pointers to CI runs demonstrating the error in a newly generated project that occurs if the change is not made.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [ ] Visual Testing
- [x] Automated Testing
- [ ] Accessibility Testing

## Checklist


packages/outline-core/src/controllers/light-dom-styles.ts
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable no-console */
 import { ReactiveControllerHost, CSSResultGroup, CSSResult } from 'lit';
-import { addScopeToStyles } from '../internal/light-dom';
+import { addScopeToStyles } from '../internal/light-dom.mjs';
```
tekante, line 3: Fixes: https://github.com/phase2/outline/actions/runs/4845713591/jobs/8634855888

packages/outline-core/src/internal/light-dom.d.mts
tekante, line 1: Fixes https://github.com/phase2/outline/actions/runs/4845876672/jobs/8635096981

packages/outline-core/tsconfig.build.json
```diff
@@ -2,7 +2,8 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "rootDir": ".",
-    "outDir": "./dist"
+    "outDir": "./dist",
+    "allowJs": true
```
tekante, line 6: Fixes https://github.com/phase2/outline/actions/runs/4845713591/jobs/8634855888

packages/outline-templates/default/rollup.config.mjs
```diff
@@ -48,6 +48,7 @@ const defaultOutput = {
       'src/components/**/*.css.lit.ts',
       'src/components/**/*.test.ts',
       'src/components/examples/**/*',
+      'src/components/sample/**/*',
```
tekante, line 51: Fixes https://github.com/phase2/outline/actions/runs/4845995503/jobs/8635294663

.github/workflows/build.yml
```diff
@@ -13,6 +13,7 @@ on:
       - beta
       - rc
       - feature/local-dev-dx
+      - feature/comprehensive-updates
```
mike-potter: I recommend that we audit this list of branches and patterns.  Doesn't seem like we should be listing specific feature branches here, but should have a pattern like `feature/*`.   If we don't want to trigger the build on every feature branch, then come up with a naming convention that is used to trigger the build.

packages/outline-templates/default/tsconfig.json
```diff
@@ -42,6 +42,7 @@
   "exclude": [
     "./src/components/**/*.stories.ts",
     "./src/components/**/*.lit.ts",
+    "./src/components/sample/**/*",
```
mike-potter, line 45: We exclude `src/components/examples/**` above in the `rollup` config, so should we also exclude them here?

.github/workflows/build.yml
tekante: True, these two are only here to facilitate development. I'll make sure to clear them before any merging.

packages/outline-templates/default/tsconfig.json
tekante: It might make sense to (or maybe examples is old and should just be removed). I'll see what others have to say to help guide the right end state.

bardleb: Do we want to ignore TS checks on samples and stories? I would think we want them regardless of their existence in the build to make sure our stories, examples, and samples are proper TS?

tekante: This was added due to [this error](https://github.com/phase2/outline/actions/runs/4846050499/jobs/8635389374) though that error may be an artifact of the local linking to facilitate testing ability to initialize a project that builds successfully. Perhaps I should only inject this into the `tsconfig.json` and `rollup.config.mjs` file during the CI process?

Or, if there is a better adjustment you are aware of that you think might allow the us to keep the files included I'm all for it (for example, should I inject the linked outline directory via https://www.typescriptlang.org/tsconfig#rootDirs)?

bardleb: I think this issue might be irrelevant to what your trying to do here. I'm fine with merging this and dealing with it later. samples shouldnt really have typescript issues apart from missing or unused imports so I think its okay. I was just raising a general question.

bardleb (approved):

PR #386, opened 2023-05-02
fix: updated component structure docs

bardleb:
## Description
- Updating Docs for component Structure


packages/outline-docs/src/guides/development/component-development/01-component-structure.mdx
```diff
@@ -0,0 +1,140 @@
...
+  static styles: CSSResultGroup = [componentStyles];
+
+  render(): TemplateResult {
+    return html`Hello World!`;
+  }
+}
+```
+## Extending Components
+### Extending a component (alias only)
+Anytime you need to use a base component, you must extend it. Base components are always named 'outline-core-componentName' by default. The extended name should simply be 'outline-componentName'. This example is the most basic form of extending which just replaces the base styles with any custom project styles you may have. Note: You will need to create a new stylesheet in your extended component directory.
```
damontgomery: Question: are the names a combination of kabob and camelcase?

`outline-my-component` or `outline-myComponent`?

The documentation implies the `outline-myComponent` to me.

damontgomery (approved):
Overall this makes sense. It seems to quickly state that this is mostly Lit, but here are a few Outline specific patterns conventions which is useful.

Small notes:
- The naming is a little unclear to me, see comment.
- The headers in this file don't seem to follow the straight hierarchy. For example, there is an h4 directly below an h2 and some of the sections seem like they are siblings in scope, but are children in the hierarchy. Like the web, IDEs can use that data to build a nice index, so maybe we could take another look.

himerus (approved):

PR #388, opened 2023-05-16
fix: Updated Outline Element

bardleb:
## Description

Removed all unnecessary code from Outline Element.


packages/outline-core/src/outline-element/outline-element.ts
```diff
@@ -1,59 +1,6 @@
...
-        </${unsafeStatic(wrapperElementType)}>`
-      : null;
-  }
-}
-
-declare global {
-  interface HTMLElementTagNameMap {
-    'outline-element': OutlineElement;
-  }
-}
```
himerus: Please retain this item for TypeScript usages. 

https://stackoverflow.com/questions/65148695/lit-element-typescript-project-global-interface-declaration-necessary

himerus (changes requested):

himerus (approved):

PR #390, opened 2023-05-17
fix: adding more notes

glitchgirl:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [X] Visual Testing
- [ ] Automated Testing
- [ ] Accessibility Testing

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

- **Documentation**
	- Introduced comprehensive documentation for component development, covering:
		- Additional resources and learning tools.
		- Coding guide and standards for building web components.
		- Component structure, composition, and extension.
		- Styling components, including guidance on CSS variables and Tailwind CSS.
		- Utilizing controllers within the Outline design system.
		- Explaining the render method and usage of partial templates.
		- Best practices for defining properties in components.
		- Using slots in web components.
		- Overview of component lifecycle methods.
		- Implementing event handlers for user interactions.
		- Utility functions for component development.
		- Extending components with Lit and Outline, including style and property overrides.


packages/outline-docs/src/guides/development/component-development/06-properties.mdx
```diff
@@ -21,7 +21,7 @@ import '@phase2/outline-container';
   bottom-margin="spacing-8"
 >
   ## Description
-  Lit components receive input and store their state as JavaScript class fields or properties.
+  Outline components receive input and store their state as JavaScript class fields(normally called slots) or properties.
```
shaal: did you mean `(normally called attributes)`?

packages/outline-docs/src/guides/development/component-development/06-properties.mdx
```diff
@@ -35,14 +35,21 @@ import '@phase2/outline-container';
...
 - Attribute names must not collide with HTML standard ones `title`, `href`, etc.
   - This means it is best to name properties more like `link-title`, `link-href`, etc.
 
+- Lit [documentation]()
+
+### Slots versus properties
+- When do I use a property? 
+  - Properties are used for non-content features that impact the state of the component. For example, active versus inactive states.
+- The component shouldn't change its own public properties, except in response to user input. For example, you could set a property to change background color, but the component itself shouldn't change it's own color.
+- Slots are used for content, such as text, icons, or images.
```
shaal:
```suggestion
- Slots are used for content, such as HTML markup, text, icons, or images. 
```

packages/outline-docs/src/guides/development/component-development/06-properties.mdx
glitchgirl: yeah i think that makes more sense

packages/outline-docs/src/guides/component-development/06-properties.mdx
```diff
@@ -21,27 +21,34 @@ import '@phase2/outline-container';
...
-  Lit components receive input and store their state as JavaScript class fields or properties.
+  Outline components receive input and store their state as JavaScript class fields(normally called attributes) or properties.
   
   [Reactive properties](https://lit.dev/docs/components/properties/) are properties that can trigger the reactive update cycle when changed, re-rendering the component, and optionally be read or written to attributes.
 
   ## Documentation Status
-  <outline-alert status="warning" size="large">
-    <span slot="header">Status: Needs Work/In Progress</span>
+  <outline-alert status="success" size="large">
+    <span slot="header">Status: Complete / Update to data May 2023</span>
```
himerus, line 30: Up to data? 😁

packages/outline-docs/src/guides/component-development/06-properties.mdx
```diff
@@ -21,27 +21,34 @@ import '@phase2/outline-container';
...
 
 ### Property usage basics
-
 - Properties should be commented using [JSDoc](https://github.com/jsdoc/jsdoc).
 - Each property should provide a comment including a description of the attribute functionality or purpose.
 - Pass to `type` the appropriate [converter](https://lit.dev/docs/components/properties/#conversion-type) (`String`, `Boolean`, etc.).
 - Pass to `attribute` an attribute name in kebab-case.
 - Attribute names must not collide with HTML standard ones `title`, `href`, etc.
   - This means it is best to name properties more like `link-title`, `link-href`, etc.
+- Lit [documentation]()
```
himerus: Missing link/info?

himerus (changes requested):
Minor adjustments.

packages/outline-docs/src/guides/component-development/06-properties.mdx
```diff
@@ -21,27 +21,34 @@ import '@phase2/outline-container';
...
+
+### Slots versus properties
+- When do I use a property? 
+  - Properties are used for non-content features that impact the state of the component. For example, active versus inactive states.
```
mkochendorfer: I don't know that I agree that content verses non-content is truly the distinguishing difference for when to use a slot verses not. There are plenty of instances in which it makes more sense to feed in plain text content via a property to make it easier to style than it would be using a slot instead. This is also the case for imagery where you often might need to apply the image as a background or do other things with it which are far more difficult if it is coming in via a slotted img tag.

packages/outline-docs/src/guides/component-development/06-properties.mdx
```diff
@@ -21,27 +21,34 @@ import '@phase2/outline-container';
...
+- The component shouldn't change its own public properties, except in response to user input. For example, you could set a property to change background color, but the component itself shouldn't change it's own color.
```
mkochendorfer: I don't know that I agree with this statement entirely. I agree that the most common case for a component to update a reflected attribute would be based on some user action, but I think there are definitely cases where a component is doing something internally that warrants reflecting its internal state back in the HTML markup regardless of any user action.

Using the background color as an example, what if your component wants to go look up the browser configuration for dark mode and modify the background color that it uses in some way? Or override the default color with a user preference setting it pulls from an API endpoint?

A more realistic example would be timed carousel with a property the reflects the currently active slide index.

I feel like a better callout might be to only use `reflect: true` on a property when the component will actually modify that property internally, which is a pattern that many components in the repo are already failing to follow correctly, and this is a mistake I see made constantly by extension on my project:
https://github.com/search?q=repo%3Aphase2%2Foutline+%22reflect%3A+true%22&type=code

https://lit.dev/docs/components/properties/#reflected-attributes

packages/outline-docs/src/guides/component-development/06-properties.mdx
```diff
@@ -21,27 +21,34 @@ import '@phase2/outline-container';
...
+- Slots are used for content, such as HTML markup, text, icons, or images.
```
mkochendorfer: Related to the above, I think this statement goes to far and also missed in my mind the actual primary use case for slots: flexible regions of a component

Icons in particular it really depends on whether the icons are built into the Outline package itself and thus are set using the name of the icon as a property on a component or if it is an svg upload that the content team does using any svg they wish.

Text often needs to be used as alt text or other things requiring attributes inside of the component, and thus cannot always use slots either.

packages/outline-docs/src/guides/component-development/06-properties.mdx
```diff
@@ -21,27 +21,34 @@ import '@phase2/outline-container';
...
 - Attribute names must not collide with HTML standard ones `title`, `href`, etc.
   - This means it is best to name properties more like `link-title`, `link-href`, etc.
+- Lit [documentation]()
+
+### Slots versus properties
+- When do I use a property? 
+  - Properties are used for non-content features that impact the state of the component. For example, active versus inactive states.
+- The component shouldn't change its own public properties, except in response to user input. For example, you could set a property to change background color, but the component itself shouldn't change it's own color.
+- Slots are used for content, such as HTML markup, text, icons, or images. 
+- There are cases in which adding content as a property will make sense, for example using a title to control the state of the component. Use this cautiously, as it can cause issues with a11y tools and automation.
```
mkochendorfer: Do we have specific examples of how this can go wrong with a11y tools? My understanding has been that that was really just a lag in the a11y tools support for web components/shadow dom, which is quickly being addressed as web component usage is exploding across the web with increased browser support.

packages/outline-docs/src/guides/component-development/06-properties.mdx
glitchgirl: added context

packages/outline-docs/src/guides/component-development/06-properties.mdx
glitchgirl: added this discussion

himerus: I agree with this. A lot of the original components, IIRC had `reflect: true` on by default as "test mode", and it was argued that it was valuable to have that. 

I agree it shouldn't be the default, but used when necessary and appropriate.

packages/outline-docs/src/guides/component-development/06-properties.mdx
himerus: > Use this cautiously, as it can cause issues with a11y tools and automation.

I think this relates to how some tools still are unable to pierce the Shadow DOM for inspecting things and may fail a11y tests if we, say build the `h1` tag inside the ShadowDOM, but the tool doesn't read it as such. 

@glitchgirl correct me if I'm wrong there in your thinking.

packages/outline-docs/src/guides/component-development/06-properties.mdx
himerus: Perhaps clarification can be given by some simple scenarios like:

For Example:
- "if you were building Web Components to be consumed by a React App that does X, Y, and Z, you might do it this way.
- "if you were building Web Components to be consumed by a Drupal App that does X, Y, and Z, you might do it this way.

I feel like that's probably where some of the assumptions are coming from. Our typical work still being Drupal, we need to account in our docs (and sample integrations would be nice) how to do things with different consumers.

packages/outline-docs/src/guides/component-development/06-properties.mdx
glitchgirl: that is exactly correct

bardleb (approved):

himerus:
Marking this as Draft, as the system is thinking these are all new files, not to be merged with the updates I had made and moving things around, etc.

PR #402, opened 2023-06-23
fix: remove lightDomStyles import from core

bardleb:
## Description

Removing the import for lightDomStyles controller as it currently prevents any components from being created by developers. 

Issues blocking projects documented here: https://phase2.slack.com/archives/C01RRSPB62F/p1687539660099819

### DIFF
--- a/packages/outline-form/src/outline-form.ts
+++ b/packages/outline-form/src/outline-form.ts
@@ -1,8 +1,8 @@
 import { TemplateResult, html, CSSResultGroup } from 'lit';
-import { OutlineElement, LightDomStyles } from '@phase2/outline-core';
+import { OutlineElement } from '@phase2/outline-core';
 import { customElement } from 'lit/decorators.js';
 import componentStyles from './outline-form.css.lit';
-import globalStyles from './outline-form.global.scoped.css.lit';
+//import globalStyles from './outline-form.global.scoped.css.lit';
 
 /**
  * The Outline Form component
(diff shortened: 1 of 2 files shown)


mike-potter:
So yes, the `outline-form` component uses the LightDomStyles, so you'd need to remove it also.

But this makes me wonder why the previous `outline-form` component WAS working and extending OutlineElement, but somehow projects can't create their own components from OutlineElement??

shaal (approved):
@bardleb thank you. Approving so you can merge it easier.
For some reason outline-form has that as dependency, and has to be removed/update there first.

packages/outline-form/src/outline-form.ts
```diff
@@ -12,9 +12,9 @@ import globalStyles from './outline-form.global.scoped.css.lit';
  */
 @customElement('outline-form')
 export class OutlineForm extends OutlineElement {
-  static styles: CSSResultGroup = [OutlineElement.styles, componentStyles];
-
-  lightDomStyles = new LightDomStyles(this, globalStyles);
+  static styles = componentStyles;
+  // Removing due to LightDomStyles dependency
+  //lightDomStyles = new LightDomStyles(this, globalStyles);
```
mike-potter, line 17: Be aware that without the light dom styles, outline-form is going to break for any project using it (will no longer have proper styling)

bardleb: That is expected.

himerus: After reviewing this carefully, the `outline-form` component is already broken: https://outline.phase2tech.com/?path=/docs/content-form--documentation

Something has changed through unlocked dependencies and a breaking change somewhere. 

Bottom line is this: 

- This is a blocker for new projects (and some current projects
- We have projects using the `lightDomStyles` controller from `@phase2/outline-core@0.1.12` (latest). 
- I can't remove that controller without it breaking this project if they update `outline-core`.

himerus:
Closed in favor of work that fixed the primary issue, and upcoming work to address the future state of this particular functionality.

PR #404, opened 2023-06-27
`outline-core-link`: Updated documentation and functionality

himerus:
## Description

Updates `outline-core-link` to be more flexible and well documented.


packages/outline-core-link/src/outline-core-link.ts
```diff
@@ -85,20 +95,87 @@ export class OutlineCoreLink extends OutlineElement {
...
+    console.groupEnd();
+  }
+
+  adjustSlottedContent(): void {
+    const slottedLink: HTMLAnchorElement | null = this.querySelector('a');
+    if (this.linkTarget) {
+      slottedLink?.setAttribute('target', this.linkTarget);
+    }
+    if (this.linkRel) {
+      slottedLink?.setAttribute('rel', this.linkRel);
```
himerus, line 155: Let's not force override. 
Let's use scenarios where we warn if things are missing on the slotting content. 

Will make follow-up for adding back in some logic for sanity checks on slotted links that may have missing or incorrect attributes.

packages/outline-core-link/src/outline-core-link.ts
```diff
@@ -85,20 +95,87 @@ export class OutlineCoreLink extends OutlineElement {
...
+  }
+
+  /**
+   * If the element is fully slotted, return the full markup in the default slot.
+   *
+   * @todo - Enable a global debug mode in outline.config.js that will determine if the console.group is logged.
+   * @returns HTMLSlotElement
+   */
+  fullMarkupInSlot(): TemplateResult {
+    const debug = true;
```
himerus: Turn to `false` at publish.

packages/outline-core-link/src/outline-core-link.ts
himerus: I don't remember the conversation that led to my prior comment. 
I'm going to allow the override to exist for now. It's documented, and can be used or not by the consumer application.

packages/outline-core-link/src/outline-core-link.ts
```diff
@@ -24,9 +28,15 @@ import componentStyles from './outline-core-link.css.lit';
...
-@customElement('outline-core-link')
+@customElement(componentName)
 export class OutlineCoreLink extends OutlineElement {
   static styles: CSSResultGroup = [componentStyles];
+
+  /**
+   * If the link is slotted, we'll test, modify & render the slotted link instead of the linkHref and linkText properties.
+   */
+  @state() isSlottedLink = false;
+
```
shaal: This seems to be an unused code.
The only other code that has `isSlottedLink`, is setting it to `true`.

packages/outline-core-link/src/outline-core-link.ts
```diff
@@ -85,20 +95,102 @@ export class OutlineCoreLink extends OutlineElement {
...
+  }
+
+  /**
+   * Adjust attributes on a slotted link.
+   *
+   * When the link is slotted, we'll test, modify & render the slotted link,
+   * and adjust the target and rel attributes if they are set on the outline-core-link element.
+   */
+  adjustSlottedContent(): void {
+    const slottedLink: HTMLAnchorElement | null = this.querySelector('a');
```
shaal:
```suggestion
    const slottedLink = this.querySelector('a') as HTMLAnchorElement | null;
```

packages/outline-core-link/src/outline-core-link.ts
```diff
@@ -85,20 +95,102 @@ export class OutlineCoreLink extends OutlineElement {
...
+    if (this.linkTarget) {
+      slottedLink?.setAttribute('target', this.linkTarget);
+    }
+    if (this.linkRel) {
+      slottedLink?.setAttribute('rel', this.linkRel);
+    }
```
shaal, line 156: I don't think outline-core-link should offer such generic override of attributes of the slot.
Alternatively, we can use something like that to provide an attribute of open new tab yes/no. When it should open in new tab, this code can set the specific attributes of rel and target.

himerus: Right now, I consider this a feature, not a bug. We can review if it's a good idea later. It's already documented, and we can choose not to utilize the ability. Up to the consumer. Flexibility.

Ability for consumer to provide an attribute that overrides something maybe supplied by default on fielded links of some sort. ¯\_(ツ)_/¯

packages/outline-core-link/src/outline-core-link.ts
```diff
@@ -85,20 +95,102 @@ export class OutlineCoreLink extends OutlineElement {
...
+  /**
+   * If the element is fully slotted, return the full markup in the default slot.
+   *
+   * @todo - Enable a global debug mode in outline.config.js that will determine if the console.group is logged.
+   * @returns HTMLSlotElement
+   */
+  fullMarkupInSlot(): TemplateResult {
+    const debug = false;
+    if (!this.isValidTopLevelSlottedLink()) {
+      if (debug) this.debugSlottedContent();
```
shaal: I think only `if (debug)` we should call `isValidTopLevelSlottedLink()`

shaal (dismissed):
I left a few comments

packages/outline-core-link/src/outline-core-link.ts
himerus: Good catch. No need for the logic going on there. Other comments made, I'll adjust this last bit.

himerus: I've made this so that the logic is more straightforward, and until the "debugController` is in place, I want to see this console error in the browser if someone is implementing the component improperly. So the debug mode is "on" always in this case.

packages/outline-core-link/src/outline-core-link.ts
himerus: If you're point is that our most basic link shouldn't do something like that, then to that I'd say again, it's likely good for now, then able to be pulled out to another location that makes more sense for this type of ability/manipulation of markup.

PR #418, opened 2023-08-22
Adding starting point for `modern` template.

himerus:
## Description

Make something other than the `default` template, which is dated, and bloated by Storybook.


himerus:
Stale.

PR #420, opened 2023-08-22
feat: implement adoptedStyleSheets + optional native CSS nesting

shaal:
## Description

This PR simplifies how we add slot styles from within the component.
* Style Manager controller does the following:
* Reads the CSS (that alternatively could be loaded globally by the consumer)
* Only if `generateRandomId === true` -
  * Generate unique ID for the component (highest specificity)
  * Formats the CSS as a string, strips `:host` and `:root`
  * Uses native CSS nesting to add the unique ID ahead of the CSS rules

* Uses `adoptedStylesheets` to modify the document's CSS (instead adding CSS into the DOM)


himerus (changes requested):
I think a lot of this is great. 

We need to, however, consider the browser impact of the Native CSS nesting. It limits us to more browsers than `adoptedStylesheet` does presently. In 6 months, maybe a year, I think we'll be there. Based on several discussions around the browser support and our ongoing requirements, it's not a safe time to implement native nesting. We still have (and will for some time) have `postcss` in our toolchain. 

If we could:

- [ ] Remove CSS Nesting
- [ ] Evaluate `addPrefixToCurlyGroups` when we aren't using native nesting
- [ ] Focus on the functionality of `addSlotStyles`

This controller should primarily be responsible for:

- [ ] Attaching a previously processed CSS file to `adoptedStylesheet`
- [ ] Optionally wrapping the CSS file in a configurable selector. (Random ID, Class, Component, etc.)

himerus:
Closing in favor of the slim [@phase2/outline-adopted-stylesheets-controller](https://www.npmjs.com/package/@phase2/outline-adopted-stylesheets-controller) controller

